PR #25, opened 2021-12-23
Fix job list not shown when not in editing mode

g-bougard:
Also reuse delete button on job list when not editing
Hide taskjobs_list when selecting a job

### DIFF
--- a/inc/taskjobview.class.php
+++ b/inc/taskjobview.class.php
@@ -159,6 +159,10 @@ class PluginGlpiinventoryTaskjobView extends PluginGlpiinventoryCommonView
     {
         $ID = $this->ajaxLoadItem($options);
         $this->showForm($ID, $options);
+       // hide taskjobs_list after a job has been selected from the taskjobs_list
+        echo Html::scriptBlock("$(document).ready(function() {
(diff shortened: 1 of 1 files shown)


inc/taskjobview.class.php
```diff
@@ -159,6 +159,10 @@ public function ajaxGetForm($options)
     {
         $ID = $this->ajaxLoadItem($options);
         $this->showForm($ID, $options);
+       // hide taskjobs_list after a job has been selected from the taskjobs_list
+        echo Html::scriptBlock("$(document).ready(function() {
+            document.getElementById(\"taskjobs_list\").style.display =\"none\";
```
orthagh, line 164:
```suggestion
           $("#taskjobs_list").hide();
```

orthagh (changes requested):

inc/taskjobview.class.php
g-bougard: Ty, I followed your suggestion in the lastest commit but protecting the quote ;-)

orthagh (approved):

trasher (approved):

PR #116, opened 2022-05-19
Drop inventory_frequency from plugin

trasher:


trasher:
Follows https://github.com/glpi-project/glpi/pull/11625

cedric-anne (approved):

PR #123, opened 2022-06-02
Migration of inventory rules

orthagh:


orthagh:
not finished at all, i must check about (maybe) truncation of existing core rules before merge

orthagh:
After discussion with partner, we agreed on:

- we change sub_type
- we preprend `[MIGRATED_FROM_FUSION]` to identify migrated rules
- we move ranking just after the max found in native rules
- we don't touch native rules (as we cannot differenciate default, user, or already migrated ones on update process)

GLPI admin  must check its rules after migration.
Maybe we can add a post migration message (can we do this with `Session::addMessageAfterRedirect` ?)

orthagh:
This is ready for review

cedric-anne (approved):

trasher (approved):

PR #137, opened 2022-06-21
fix(job): fix empty job log

stonebuzz:
Before (nothing display)

image

After

image

### DIFF
--- a/inc/task.class.php
+++ b/inc/task.class.php
@@ -202,19 +202,7 @@ class PluginGlpiinventoryTask extends PluginGlpiinventoryTaskView
     }
 
 
-   /**
-    * Get standard massive action forbidden (hide in massive action list)
-    *
-    * @return array
-    */
-    public function getForbiddenStandardMassiveAction()
-    {
-        $forbidden = parent::getForbiddenStandardMassiveAction();
-        $forbidden[] = 'update';
-        return $forbidden;
-    }
-
-       /**
+    /**
     * Display form
     *
     * @param object $item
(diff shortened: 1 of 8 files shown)


cedric-anne (approved):

trasher (approved):

g-bougard (approved):

g-bougard (changes requested):
Finally, it seems to have a side-effect on the tasks shown from the ip range entries: all tasks are shown there where only tasks using the ip range should be.

inc/task.class.php
```diff
@@ -166,6 +166,100 @@ public function rawSearchOptions()
...
+    /**
+    * Get the tab name used for item
+    *
+    * @param object $item the item object
+    * @param integer $withtemplate 1 if is a template form
+    * @return string name of the tab
+    */
+    public function getTabNameForItem(CommonGLPI $item, $withtemplate = 0)
+    {
+
```
g-bougard, line 178: It seems this blank line makes lint fail.
```suggestion
```

g-bougard (approved):

trasher (approved):

inc/task.class.php
```diff
@@ -166,6 +166,99 @@ public function rawSearchOptions()
...
+    * Get the tab name used for item
+    *
+    * @param object $item the item object
+    * @param integer $withtemplate 1 if is a template form
+    * @return string name of the tab
+    */
+    public function getTabNameForItem(CommonGLPI $item, $withtemplate = 0)
+    {
+        if ($item->fields['id'] > 0) {
+
```
trasher, line 179:
```suggestion
```

orthagh (approved):

PR #206, opened 2022-08-19
feat(task job log): add logs when prepare did not return any target

stonebuzz:
Add more log when SNMP inventory task (preparation step) did not return any target

image

### DIFF
--- a/js/taskjobs.js
+++ b/js/taskjobs.js
@@ -710,6 +710,11 @@ taskjobs.update_logs = function (data) {
          targets_selector = job_selector + ' .targets_block';
 
          var targets_cpt = 0;
+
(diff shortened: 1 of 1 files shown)


js/taskjobs.js
```diff
@@ -710,6 +710,11 @@ taskjobs.update_logs = function (data) {
          targets_selector = job_selector + ' .targets_block';
 
          var targets_cpt = 0;
+
+         if(job_v.targets.length === 0) {
+            $(targets_selector).html('The preparation of the task did not return any targets');
```
trasher, line 715: This will not be translated (note I do not know hos to translate JS messages)

trasher (approved):

PR #249, opened 2022-11-08
remove useless right

stonebuzz:
Remove useless right that have been migrated to core or useless now
see !25502


trasher:
I guess we can remove instead of commenting

stonebuzz:
> I guess we can remove instead of commenting

other unnecessary rights have just been commented on.

practical if you have to come back to it (if needed)

trasher (approved):

cedric-anne (approved):

Issue #259, opened 2022-11-25
Wrong IP range boxes disposition in edition form


g-bougard:
### Describe the bug

IP range numbers are splited into dedicated boxes and they can be put on a new line.

### To reproduce

Edit a created IP range. The following case in on a 1920x1080 monitor.
![image](https://user-images.githubusercontent.com/12514489/203977655-47281f2a-c54a-4256-9f35-d5bd341643a7.png)
You can see a lot of space is lost and the last number of ip ranges goes on a new line.

You even can have the last 2 boxes on a new line with a less large window:
![image](https://user-images.githubusercontent.com/12514489/203978414-3e9bb888-9934-4499-899e-031a20a8bfd8.png)

### Expected behavior

IP numbers are not splited othe lines.

### Operating system

Linux

### GLPI Agent version

1.4

### GLPI version

10.0.5

### GLPIInventory plugin

1.0.5

stonebuzz:
hi @g-bougard , thank you for your feedback

Can you try this -> https://github.com/glpi-project/glpi-inventory-plugin/pull/262

Best regards

PR #262, opened 2022-11-25
fix(iprange): fix input size to prevent bad display

stonebuzz:
fix input size to prevent bad display

Before  : 
image

After  : 
image

Fix : #259

### DIFF
--- a/templates/forms/iprange.html.twig
+++ b/templates/forms/iprange.html.twig
@@ -94,19 +94,19 @@
       <div class="row">
          <div class="col-auto">
             <input type="text" class="form-control" value="{{ ipexploded[0] }}"
-                   name="{{ name }}0" id="{{ name }}0" data-index="0" size="3" maxlength="3">
+                   name="{{ name }}0" id="{{ name }}0" data-index="0" size="1" maxlength="3">
          </div>
          <div class="col-auto">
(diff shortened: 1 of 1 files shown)


orthagh:
maybe adding `.flex-nowrap` to the row should fix better

stonebuzz:
> maybe adding `.flex-nowrap` to the row should fix better

add ```flex-nowrap``` class and re reduces the size a little bit (in order not to exceed the ```entity``` input) 

image

g-bougard:
But something is still wrong: IP boxes overlaps when you stretch the window
image

stonebuzz:
what is the size of your screen

cedric-anne:
> what is the size of your screen

size does not matters

stonebuzz:
> > what is the size of your screen
> 
> size does not matters

resolution ?

stonebuzz:
it is because of the ```flex-nowrap``` on small resolution

stonebuzz:
@g-bougard 

can you retry ?

stonebuzz:
Can you retry ?

g-bougard:
Seems good this way.

g-bougard (approved):

cedric-anne (approved):

PR #277, opened 2022-12-13
fix(deploy): fix modal and add infos

stonebuzz:
fix modal and add infos

Before : 
image

After : 
image

### DIFF
--- a/inc/deployfile.class.php
+++ b/inc/deployfile.class.php
@@ -375,7 +375,7 @@ class PluginGlpiinventoryDeployFile extends PluginGlpiinventoryDeployPackageItem
          enableDD         : true,
          containerScroll  : true,
          height           : 320,
-         width            : 770,
+         width            : 670,
          loader           : Tree_Category_Loader$rand,
          rootVisible      : false,
          listeners: {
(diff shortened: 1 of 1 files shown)


inc/deployfile.class.php
```diff
@@ -403,6 +403,11 @@ public static function showServerFileTree($rand)
 
         echo "</script>";
 
+        echo '<div class="alert alert-info d-flex" role="alert">';
+        echo '<i class="fas fa-exclamation-circle fa-2x me-2"></i>';
+        echo '<p>' . __('Files need to be uploaded to <kbd>/files/_plugins/glpiinventory/upload/</kbd> folder to be displayed here', 'glpiinventory') . '</p>';
```
cedric-anne, line 408: `<kbd>` is to hint about keyboard input, you should use another HTML tag.

stonebuzz: image

g-bougard (approved):

inc/deployfile.class.php
```diff
@@ -403,6 +403,11 @@ public static function showServerFileTree($rand)
 
         echo "</script>";
 
+        echo '<div class="alert alert-info d-flex" role="alert">';
+        echo '<i class="fas fa-exclamation-circle fa-2x me-2"></i>';
+        echo '<p>' . __('Files need to be uploaded to <mark>/files/_plugins/glpiinventory/upload/</mark> folder to be displayed here', 'glpiinventory') . '</p>';
```
trasher, line 408: I guess tags and path should not be part of translated string
```suggestion
        echo '<p>' . sprintf(__('Files need to be uploaded to %s folder to be displayed here', 'glpiinventory'), '<mark>/files/_plugins/glpiinventory/upload/</mark>') . '</p>';
```

trasher (changes requested):

PR #306, opened 2023-02-02
feat(task): prepare target with remote_addr if available

stonebuzz:
Prepare change for https://github.com/glpi-project/glpi/pull/13935

When ```remote_addr``` field is available (from ```NetworkEquipement``` / ```Printer```) use it to find target for ```netinventory```


stonebuzz:
I close this PR, we will come back on the subject the day of implementation of the global network inventory (without network discovery)

Issue #308, opened 2023-02-09
Static Groups messed up


grego1981:
### Describe the bug

Coming from 9.5 and FI all static groups are not working as they should. Cannot remove a PC from the group and in Assosiated items, while the count in the menu is correct inside appear computers from other groups.

### To reproduce

Go to GLPI Inventory->General-> Groups of computers, select a group and go to assosiated items
Baloon with item count is correct but in the list appear items that do not belong to that group

### Expected behavior

Items visisble should be only the ones belong to that group.
There should be an action to remove assosiation to the group.

### Operating system

Windows, Linux

### GLPI Agent version

1.4

### GLPI version

10.0.5

### GLPIInventory plugin

1.1.0

stonebuzz:
Hi @grego1981 
can you try this -> https://github.com/glpi-project/glpi-inventory-plugin/pull/309

grego1981:
After applying the 4 files in the fix I can see the remove from group. The issue remains for the view where appear not only the assosiated computers but from other groups as well... Also if trying to sort by any column the view changes and returns ALL computers in a view containg only computername entity name and the field you selected to sort. though the plugin's field static group does not appear at all, not even if you try to select it in the view settings. (wrench  icon)

stonebuzz:
Hi @grego1981
 
can you retry ?

(need to logout / login because of  latered ```$_SESSION```)

grego1981:
I've tested, now column sorting does not work at all. Adding and removing works. Also wrong machines still appear in the assosiated items..

stonebuzz:
Sorting feature have been drop from this two forms (to prevent refresh and wrong display).

"Associated items" tab have been rework entirely

![image](https://user-images.githubusercontent.com/7335054/218482322-07ef8913-0ec3-451c-ad09-f98883ff1874.png)

Are display only computer linked to static group (from database)

grego1981:
You mean that this is what I should see or is it a new addition for the new version? What I'm seeing in the assosiated items is the view I have in the computers view, and not the assosiated to the group computers only..

stonebuzz:
I just say that this view has been reworked, both were based on data in database, but the old view used the framework of GLPI to display the table (which caused display problems when sorting for example).

Now it is the plugin that builds the table

[...]

PR #309, opened 2023-02-09
fix(deploygroup_static): fix and harmonize massive action

stonebuzz:
Actually specific massives actions ```add to group``` or ```remove from group``` are not displayed

image

Now : 

From ```criteria``` tab we can add computer to group easily
image

From ```Associated elements``` tab we can delete computer to group easily
image

This PR try to harmonize massives actions from computer by adding ```exclude_from_group```  (```add_to_group``` already present```)

image

fix : https://github.com/glpi-project/glpi-inventory-plugin/issues/308

### DIFF
--- a/inc/computer.class.php
+++ b/inc/computer.class.php
@@ -3,7 +3,7 @@
 /**
  * ---------------------------------------------------------------------
  * GLPI Inventory Plugin
- * Copyright (C) 2021 Teclib' and contributors.
+ * Copyright (C) 2021-2023 Teclib' and contributors.
  *
  * http://glpi-project.org
  *
--- a/inc/computer.class.php
+++ b/inc/computer.class.php
@@ -8,7 +8,7 @@
  * http://glpi-project.org
  *
  * based on FusionInventory for GLPI
- * Copyright (C) 2010-2021 by the FusionInventory Development Team.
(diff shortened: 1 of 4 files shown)


trasher:
This is a part I absolutely do not know... I'm not fan of override of computer, but it seems it's been done that way in the original plugin?
Also, the new file license header is wrong

inc/computer.class.php
```diff
@@ -0,0 +1,180 @@
+<?php
+
+/**
+ * ---------------------------------------------------------------------
+ * GLPI Inventory Plugin
+ * Copyright (C) 2021 Teclib' and contributors.
```
trasher, line 6:
```suggestion
 * Copyright (C) 2021-2023 Teclib' and contributors.
```

inc/computer.class.php
```diff
@@ -0,0 +1,180 @@
...
+ *
+ * http://glpi-project.org
+ *
+ * based on FusionInventory for GLPI
+ * Copyright (C) 2010-2021 by the FusionInventory Development Team.
```
trasher, line 11:
```suggestion
 * Copyright (C) 2010-2023 by the FusionInventory Development Team.
```

trasher (approved):
Seems correct, but I cannot test that

cedric-anne (approved):
Same for me.

PR #317, opened 2023-03-01
fix(tu): add where clause on itemtype

stonebuzz:
Fix TU because of  -> https://github.com/glpi-project/glpi/pull/14143

Now ```Unmanaged``` have same ```logicial_number``` for it's ```NetworkPort``` as the ```NetworkEquipment``` it comes from


stonebuzz:
close in favor of https://github.com/glpi-project/glpi-inventory-plugin/pull/316

PR #336, opened 2023-04-03
Fix PHP 8.2 deprecated notation

trasher:


trasher:
Noticed while trying to understand transfers issue

PR #362, opened 2023-05-30
fix(migration): handle lost value last_boot, remote_addr, last_inventory_update

stonebuzz:
Some data are lost during migration

- ```last_boot```
- ```last_inventory_update```
- ```remote_addr```

This PR prevent this

### DIFF
--- a/install/update.php
+++ b/install/update.php
@@ -3104,12 +3104,16 @@ function do_biosascomponentmigration()
         $DB->tableExists('glpi_plugin_glpiinventory_inventorycomputercomputers') &&
         $DB->fieldExists('glpi_plugin_glpiinventory_inventorycomputercomputers', 'remote_addr')
     ) {
-        //retrieve exiting
-        $query = "SELECT computers_id, remote_addr
-                    FROM glpi_plugin_glpiinventory_inventorycomputercomputers";
-        $result = $DB->query($query);
 
-        while ($data = $DB->fetchArray($result)) {
+        $computers_iterator = $DB->request([
+            'SELECT' => [
+                'computers_id',
+                'remote_addr'
+            ],
+            'FROM' => 'glpi_plugin_glpiinventory_inventorycomputercomputers'
+        ]);
+
+        foreach ($computers_iterator as $data) {
             $computer = new Computer();
             if ($computer->getFromDB($data['computers_id'])) {
                 $agent = $computer->getInventoryAgent();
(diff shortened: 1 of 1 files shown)


install/update.php
```diff
@@ -3099,6 +3099,69 @@ function do_biosascomponentmigration()
 {
     global $DB;
 
+    //Agent data migration -> remote_addr
+    if (
+        $DB->tableExists('glpi_plugin_glpiinventory_inventorycomputercomputers') &&
+        $DB->fieldExists('glpi_plugin_glpiinventory_inventorycomputercomputers', 'remote_addr')
+    ) {
+        //retrieve exiting
+        $query = "SELECT computers_id, remote_addr
```
trasher, line 3108: Please avoid raw SQL queries

trasher (changes requested):
See comment

install/update.php
stonebuzz: done !

stonebuzz:
> See comment

done

install/update.php
```diff
@@ -3099,6 +3099,62 @@ function do_biosascomponentmigration()
...
+                    $input_computer['last_inventory_update'] = $data['last_inventory_update'];
+                }
+
+                if (count($input_computer) > 1) {
+                    $computer->update($input_computer);
+                }
+
+                //handle remote_addr
+                $agent = $computer->getInventoryAgent();
+                if (!is_null($agent) && !empty($data['remote_addr'])) {
```
trasher, line 3144:
```suggestion
                if (!is_null($agent) && !empty($data['remote_addr']) && empty($agent->fields['remote_addr'])) {
```

install/update.php
```diff
@@ -3099,6 +3099,62 @@ function do_biosascomponentmigration()
...
+                'last_boot',
+                'last_inventory_update'
+            ],
+            'FROM' => 'glpi_plugin_glpiinventory_inventorycomputercomputers'
+        ]);
+
+        foreach ($computers_iterator as $data) {
+            $computer = new Computer();
+            if ($computer->getFromDB($data['computers_id'])) {
+
```
trasher, line 3124:
```suggestion
```

cedric-anne (commented):
There is already something that is handling `last_inventory_update` in `update.native.php`.

For instance:
```php
        $DB->queryOrDie(
            "UPDATE `glpi_computers` AS `computers`
            INNER JOIN (
              SELECT
                `computers_id`,
                `last_inventory_update`
              FROM `glpi_plugin_glpiinventory_inventorycomputercomputers`
          ) AS `plugin_computers` ON `plugin_computers`.`computers_id` = `computers`.`id`
          SET
              `computers`.`last_inventory_update` = `plugin_computers`.`last_inventory_update`
          ;"
        );
```

PR #373, opened 2023-06-28
fix(migration): rename column to prevent lost data

### LINKED ISSUE #372: Lasdt inventory date not migrated
### Describe the bug

Seems like "last inventory update" is not migrated when updating from FusionInventory plugin; see https://github.com/glpi-project/glpi/issues/14735

Note that FI plugin is not compatible with GLPI > 10.0.6; therefore I'm not 100% sure of the results of my tests (I have no time to try with old versions right now).

### To reproduce

1. Install FI plugin, do an inventory
2. Install glpiinventory plugin
3. Update plugin
4. See dates in computer main form.

### Expected behavior

Dates should al be migrated.

stonebuzz:
When ```migration``` handle ```glpi_plugin_fusioninventory_inventorycomputercomputers```

Table is renamed to ```glpi_plugin_glpiinventory_inventorycomputercomputers```

But column ```last_fusioninventory_update``` is not renamed to ```last_inventory_update```

```migration``` script create ```last_inventory_update``` (with ```NULL```) and drop ```last_fusioninventory_update``` 
instead of rename it (and keep data).

This PR fix this and fix #372 too

### DIFF
--- a/install/update.php
+++ b/install/update.php
@@ -3052,7 +3052,7 @@ function do_computercomputer_migration($migration)
       'bios_manufacturers_id'
     ];
 
-    $a_table['renamefields'] = [];
+    $a_table['renamefields']['last_fusioninventory_update'] = 'last_inventory_update';
 
     $a_table['keys']   = [];
     $a_table['keys'][] = ['field' => 'computers_id', 'name' => '', 'type' => 'INDEX'];


trasher:
ping @AldarisPale

AldarisPale:
Thanks, checking

AldarisPale:
@stonebuzz looks like it's working now! Thank you very much!
I can see that the following fields have been migrated: 
* Last contact
* **Last inventory update**
* Last boot date

trasher (approved):

Issue #375, opened 2023-07-11
 GLPI Inventory 1.2.2 : Error sql when i click on the task automatic "cleantaskjob"


steph461:
### Describe the bug

hello,

I have errors that repeat randomly
this task "cleantaskjob"generate this errors
When I click execute, the task cleantaskjob generate errors
![image](https://github.com/glpi-project/glpi-inventory-plugin/assets/68054991/c1b6b728-2d74-422a-aa74-dd7c63b3720f)

 ```sheel
[2023-07-09 19:27:02] glpisqllog.WARNING: DBmysql::query() in /var/www/html/glpi/src/DBmysql.php line 419
  *** MySQL query warnings:
  SQL: SELECT * FROM `glpi_plugin_glpiinventory_taskjoblogs` WHERE `state` IN ('finished', 'cancelled') AND `date` < DATE_ADD(NOW(), INTERVAL -1 DAY) GROUP BY `plugin_glpiinventory_taskjobstates_id`
  Warnings: 
1292: Truncated incorrect DECIMAL value: 'finished'
1292: Truncated incorrect DECIMAL value: 'cancelled'
  Backtrace :
  src/DBmysqlIterator.php:112                        
  src/DBmysql.php:1078                               DBmysqlIterator->execute()
  ...ce/glpiinventory/inc/taskjobstate.class.php:573 DBmysql->request()
  src/CronTask.php:1018                              PluginGlpiinventoryTaskjobstate::cronCleantaskjob()
  front/cron.php:83                                  CronTask::launch()
  {"user":"@srv-glpi"} 
[2023-07-11 10:49:01] glpisqllog.WARNING: DBmysql::query() in /var/www/html/glpi/src/DBmysql.php line 419
  *** MySQL query warnings:
  SQL: SELECT * FROM `glpi_plugin_glpiinventory_taskjoblogs` WHERE `state` IN ('finished', 'cancelled') AND `date` < DATE_ADD(NOW(), INTERVAL -1 DAY) GROUP BY `plugin_glpiinventory_taskjobstates_id`
  Warnings: 
1292: Truncated incorrect DECIMAL value: 'finished'
1292: Truncated incorrect DECIMAL value: 'cancelled'
  Backtrace :
  src/DBmysqlIterator.php:112                        
  src/DBmysql.php:1078                               DBmysqlIterator->execute()
  ...ce/glpiinventory/inc/taskjobstate.class.php:573 DBmysql->request()
[...]

PR #376, opened 2023-07-11
fix(core): fix SQL query

stonebuzz:
Fix SQL query

fix : #375


trasher:
Seems like the state condition was not present before migration to iterator

inc/taskjobstate.class.php
```diff
@@ -567,7 +567,7 @@ public static function cronCleantaskjob()
         $iterator = $DB->request([
             'FROM'   => 'glpi_plugin_glpiinventory_taskjoblogs',
             'WHERE'  => [
-                'state' => ['finished', 'cancelled'],
+                'state' => [self::FINISHED, self::CANCELLED],
```
trasher, line 570:
```suggestion
```

cedric-anne (approved):

trasher (approved):

PR #378, opened 2023-07-12
Remove specific search class

trasher:


inc/deploygroup_staticdata.class.php
```diff
@@ -156,26 +156,24 @@ public static function showCriteriaAndSearch(PluginGlpiinventoryDeployGroup $ite
...
         $search_params['massiveactionparams']['extraparams']['specific_actions']['PluginGlpiinventoryComputer' . MassiveAction::CLASS_ACTION_SEPARATOR . 'add'] = __('Add to static group', 'glpiinventory');
         $search_params['massiveactionparams']['extraparams']['massive_action_fields'] = ['action', 'id'];
 
         $data = Search::prepareDatasForSearch('PluginGlpiinventoryComputer', $search_params);
         $data['itemtype'] = 'Computer';
         Search::constructSQL($data);
-
-       // Use our specific constructDatas function rather than Glpi function
-        PluginGlpiinventorySearch::constructDatas($data);
+        Search::constructDatas($data);
```
stonebuzz, line 176:
```suggestion
        Search::constructData($data);
```

to prevent warning

image

```Search::constructData($data);``` instead of ```Search::constructDatas($data);```

**Review by stonebuzz — CHANGES_REQUESTED** [review]

File: inc/wakeonlan.class.php [hunk 2]
```diff
@@ -125,7 +125,7 @@ public function prepareRun($taskjobs_id)
                                 $_GET["glpisearchcount2"] = count($_GET['field2']);
                             }
 
-                            $pfSearch = new PluginGlpiinventorySearch();
+                            $pfSearch = new Search();
                             Search::manageGetValues('Computer');
```
stonebuzz, line 129:
```suggestion

```
As this function does not exist (even in GLPI 9.5 and FI 9.5) perhaps we can remove it (return statement are never used)

stonebuzz (changes requested):

inc/wakeonlan.class.php
trasher: Indeed we could, but anyway this class is broken for ages... I'm preparing a PR to remove it entirely, so we can just ignore those issues here.

stonebuzz (approved):

cedric-anne (approved):

PR #398, opened 2023-09-05
remove deprecated import[rel] (was a hack anywary) to wakeup local agent

orthagh:
CF !29180

Import[rel] is deprecated in modern browsers, but anyway it was a hack.
With this change, the browser will say there is cors issue, but the agent will still get the order.

@g-bougard is currently working on adding cors headers on agent to remove the exception side of GLPI.

### DIFF
--- a/front/deploypackage.public.php
+++ b/front/deploypackage.public.php
@@ -70,7 +70,7 @@ if (isset($_POST['prepareinstall'])) {
                 $port = (int)$agent->fields['port'];
             }
             if ($port == 0) {
-                $port = 62354;
+                $port = Agent::DEFAULT_PORT;
             }
             echo Html::scriptBlock("
                 $.get('http://127.0.0.1:{$port}/now');
--- a/front/deploypackage.public.php
+++ b/front/deploypackage.public.php
@@ -69,6 +69,9 @@ if (isset($_POST['prepareinstall'])) {
                 $agent->getFromDBByCrit(['itemtype' => 'Computer', 'items_id' => $computers_id]);
                 $port = (int)$agent->fields['port'];
             }
+            if ($port == 0) {
+                $port = 62354;
+            }
             echo Html::scriptBlock("
                 $.get('http://127.0.0.1:{$port}/now');
                 setTimeout(function(){
--- a/front/deploypackage.public.php
+++ b/front/deploypackage.public.php
@@ -63,8 +63,14 @@ if (isset($_POST['prepareinstall'])) {
    //If it's a local wakeup, local call to the agent RPC service
     switch ($_POST['wakeup_type']) {
         case 'local':
+            $port = 62354;
+            if ($computers_id) {
+                $agent = new Agent();
+                $agent->getFromDBByCrit(['itemtype' => 'Computer', 'items_id' => $computers_id]);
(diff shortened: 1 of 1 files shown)


front/deploypackage.public.php
```diff
@@ -63,10 +63,12 @@
    //If it's a local wakeup, local call to the agent RPC service
     switch ($_POST['wakeup_type']) {
         case 'local':
-            echo '<link rel="import" href="http://127.0.0.1:62354/now">';
-            echo Html::scriptBlock("setTimeout(function(){
-            window.location='{$_SERVER['HTTP_REFERER']}';
-         }, 500);");
+            echo Html::scriptBlock("
+                $.get('http://127.0.0.1:62354/now');
```
g-bougard, line 67: Agent port should if possible retrieved from the agent object.

trasher: I see no way to find the agent

orthagh: Will be hard to achieve in this 'local' case as it doesn't target a specific computer, but the localhost only.
I need to find the good computer for the current user, and if he has more than one, I will not be able to find the good one.

The better I can do is retrieving the default agent port.

g-bougard: Lines 53 to 60, I see a loop which purpose seems to retrieve the `$computers_id`. Is really there no way to retrieve the related agent from this id ? And if unknown, yes, using the default configured agent port would be nice.

g-bougard: An even, there's an example in the "remote" case. For me something like this works:
```php
            if ($computers_id) {
                $agent = new Agent();
                $agent->getFromDBByCrit(['itemtype' => 'Computer', 'items_id' => $computers_id]);
                $port = (int)$agent->fields['port'];
            }
            echo Html::scriptBlock("
                $.get('http://127.0.0.1:".(isset($port) && $port > 0 ? $port : 62354)."/now');
                setTimeout(function(){
                    window.location='{$_SERVER['HTTP_REFERER']}';
                }, 500);
            ");
```

g-bougard (approved):

front/deploypackage.public.php
```diff
@@ -63,10 +63,21 @@
    //If it's a local wakeup, local call to the agent RPC service
     switch ($_POST['wakeup_type']) {
         case 'local':
-            echo '<link rel="import" href="http://127.0.0.1:62354/now">';
-            echo Html::scriptBlock("setTimeout(function(){
-            window.location='{$_SERVER['HTTP_REFERER']}';
-         }, 500);");
+            $port = 62354;
```
trasher, line 66:
```suggestion
            $port = Agent::DEFAULT_PORT;
```

front/deploypackage.public.php
```diff
@@ -63,10 +63,21 @@
...
+            if ($computers_id) {
+                $agent = new Agent();
+                $agent->getFromDBByCrit(['itemtype' => 'Computer', 'items_id' => $computers_id]);
+                $port = (int)$agent->fields['port'];
+            }
+            if ($port == 0) {
+                $port = 62354;
```
trasher, line 73:
```suggestion
                $port = Agent::DEFAULT_PORT;
```

trasher (approved):

PR #402, opened 2023-09-08
fix(netinventory): handle XML with <ERROR> node

stonebuzz:
Replace #401 

Try to handle ```<ERROR>``` node from different place

```$a_CONTENT->content->error```

already manage : 

```$a_CONTENT->content->device->error```

need https://github.com/glpi-project/inventory_format/pull/121


trasher (approved):

cedric-anne (approved):

g-bougard:
I managed to reproduce the case and with that fix it works as expected and the task log reports the problem:
image

g-bougard (approved):

cedric-anne (approved):

trasher (approved):

PR #427, opened 2023-10-11
Enable GLPI strict depreciations

trasher:
Once merged, plugin will be compatible GLPI >= 10.0.11 :exclamation: :skull_and_crossbones:


trasher:
Tehre are remainin `queryOrDie` that should also be changed

trasher:
Since this requires GLPI 10.0.1; it should not be merged too early.

cedric-anne (approved):

PR #431, opened 2023-10-31
fix duplicate entry on update

Rom1-B:
When updating the plugin, this kind of error appears, blocking the update.

```
SQL Error "1062": Duplicate entry '6-2022-04-28' for key 'unicity' in query "INSERT INTO `glpi_printerlogs` (...
```

Internal ref: !30253

### DIFF
--- a/install/update.native.php
+++ b/install/update.native.php
@@ -385,8 +385,8 @@ function pluginGlpiinventoryUpdateNative($current_version, $migrationname = 'Mig
 
     $migration->displayMessage("Use core printer logs");
     if ($DB->tableExists('glpi_plugin_glpiinventory_printerlogs')) {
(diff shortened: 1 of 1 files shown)


trasher:
I guess doQuery changes will conflict with the already opened PR.

Rom1-B:
Are you talking about https://github.com/glpi-project/glpi-inventory-plugin/pull/427?

It seems that only `query` is replaced, not `queryOrDie`

trasher:
> It seems that only `query` is replaced, not `queryOrDie`

Indeed

cedric-anne (approved):
`INSERT IGNORE` will probably emit warnings, but I guess it will only be for edge cases, so it should be OK.

trasher (approved):

PR #452, opened 2023-12-07
Fix(RuleCollect): Drop RuleCollect

stonebuzz:
Drop of "Computer information rules"

(with criteria based on collect)

They are no longer (and cannot be) called up by the native inventory without lot of work

No feedback (once) on their non-operation


trasher:
> `inc/collectrule.class.php` should have been removed aswell, no?

?

trasher (approved):

cedric-anne (approved):

PR #488, opened 2024-03-07
fix: dynamic groups missing on computer

Rom1-B:
Compute more often `computers_id_cache`, which displays dynamic groups in a computer's "Tasks / Groups" tab

ref !31924


trasher (commented):
Waiting for customer feedback, I have no idea if this is or not correct.

Rom1-B:
Customer has just validated the patch

trasher (approved):

PR #499, opened 2024-04-12
fix: cancelled count

Rom1-B:
Cancelled also displayed other statuses

Before:
image001

### DIFF
--- a/inc/task.class.php
+++ b/inc/task.class.php
@@ -1124,6 +1124,9 @@ class PluginGlpiinventoryTask extends PluginGlpiinventoryTaskView
                         if (
                             !isset($counters['agents_prepared'][$agent_id])
                             && !isset($counters['agents_running'][$agent_id])
+                            && !isset($counters['agents_success'][$agent_id])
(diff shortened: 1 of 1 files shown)


trasher:
Tests are failing

trasher (approved):

inc/task.class.php
```diff
@@ -1210,6 +1213,7 @@ public function getJoblogs($task_ids = [], $with_logs = true, $only_active = fal
                 if (
                     !isset($counters['agents_error'][$agent_id])
                     && !isset($counters['agents_success'][$agent_id])
+                    && !isset($counters['agents_cancelled'][$agent_id])
```
stonebuzz, line 1216: Shouldn't the conditions be 'all' the same in unit tests?

stonebuzz (changes requested):

inc/task.class.php
stonebuzz: after discussion, forget my previous review

stonebuzz (approved):

PR #502, opened 2024-04-16
Remove the Not done yet status

Rom1-B:
"Not done yet" is supposed to group together "Prepared" and "Running" statuses. I think it's a redundant feature, and it doesn't seem to work (I'm not sure it ever did). So I suggest simply removing it.


trasher:
I have no idea if that's been used/working once :(

stonebuzz:
After talking about it, it's a redundant feature (if it worked, but it doesn't).

the risk is not very high (and can be easily reverted if necessary, which I doubt)

stonebuzz (approved):

trasher (approved):

PR #517, opened 2024-06-10
Fix management of the “include old jobs” parameter

RomainLvr:
The `inculde old jobs` parameter only modified the number of rows generated in the csv, not the number of jobs included.
Here's a screenshot of an export with `include old jobs` set to 2: 

image

image

By increasing this value, additional lines are added : 

image

The solution was to move the counter that stops the addition of rows in the CSV, so that it applies to jobs and not rows.


Rom1-B (approved):
Seems OK

stonebuzz (approved):

Issue #523, opened 2024-07-09
 Management device IP returned to NetInventory task doesn't reflect the ones on IP range 


eduardomozart:
### Describe the bug

I created a NetInventory task to NetInventory the network range 192.168.0.0/24 but it seems that the first IP (172.31.98.1) is returned to the task:

![image](https://github.com/glpi-project/glpi-inventory-plugin/assets/2974895/9e8e622a-7ace-40cc-875a-2979ec455cec)

Print of Network devices > [Device] > Network ports.

GLPI Agent Debug (Level 1) output:

```
[Mon Jul  8 16:03:39 2024][debug] new thread 12 to handle task NetInventory
[Mon Jul  8 16:03:39 2024][info] running task NetInventory
[Mon Jul  8 16:03:39 2024][debug] Current netinventory run expiration timeout: 1.0 hour
[Mon Jul  8 16:03:39 2024][debug] using 4 netinventory workers
[Mon Jul  8 16:03:40 2024][debug] #1, full snmp scan of 192.168.0.1 with credentials 2
[Mon Jul  8 16:03:40 2024][debug] #2, full snmp scan of 172.31.98.1 with credentials 2
[Mon Jul  8 16:03:41 2024][debug] #3, full snmp scan of 172.31.98.1 with credentials 2
[Mon Jul  8 16:03:41 2024][debug] #4, full snmp scan of 172.31.98.1 with credentials 2
[Mon Jul  8 16:03:55 2024][error] #1, SNMP communication error: no response from host 192.168.0.1
[Mon Jul  8 16:03:56 2024][error] #2, SNMP communication error: no response from host 172.31.98.1
[Mon Jul  8 16:03:56 2024][error] #3, SNMP communication error: no response from host 172.31.98.1
[Mon Jul  8 16:03:56 2024][error] #4, SNMP communication error: no response from host 172.31.98.1
```

It seems to be a known glpi-inventory-plugin issue as reported by @g-bougard in https://github.com/glpi-project/glpi-agent/issues/705

Downloads:

GLPI Inventory file: [networkequipment_0_4.xml.txt](https://github.com/user-attachments/files/16134698/networkequipment_0_4.xml.txt)

### To reproduce

1. Create a new NetInventory task after a NetDiscovery task.
2. See error on GLPI Agent debug log.

### Expected behavior

Use the IP that matches the IP range setup.
OR
If the contact with the device fails through the first IP, try the next one until the contact succeeds.

### Operating system

[...]

PR #524, opened 2024-07-09
Use IP range set on task for Device NetInventory

eduardomozart:
Fix https://github.com/glpi-project/glpi-inventory-plugin/issues/523


stonebuzz (approved):
Hi @eduardomozart 

sounds OK to me (good job)

I don't have the means to test it, but the change doesn't seem risky.

Now that the problem has been clearly identified, we can revert if necessary

inc/networkinventory.class.php
```diff
@@ -367,11 +367,33 @@ public function run($jobstate)
         $current = $jobstate;
         $agent->getFromDB($current->fields['agents_id']);
 
-        $ip = current(PluginGlpiinventoryToolbox::getIPforDevice(
+        $ips = current(PluginGlpiinventoryToolbox::getIPforDevice(
             $jobstate->fields['itemtype'],
             $jobstate->fields['items_id']
         ));
```
cedric-anne, line 373: The loop `foreach ($ips as &$ip) {` will not work if you get the current element here.
```suggestion
        $device_ips = PluginGlpiinventoryToolbox::getIPforDevice(
            $jobstate->fields['itemtype'],
            $jobstate->fields['items_id']
        );
        $ip = current($ips);
```

inc/networkinventory.class.php
```diff
@@ -367,11 +367,33 @@ public function run($jobstate)
...
+        foreach ($a_taskjobs as $a_taskjob) {
+            $a_definition = importArrayFromDB($a_taskjob['targets']);
+            foreach ($a_definition as $datas) {
+                $itemtype = key($datas);
+                $items_id = current($datas);
+    
+                switch ($itemtype) {
+                    case 'PluginGlpiinventoryIPRange':
+                        $pfIPRange->getFromDB($items_id);
+                        foreach ($ips as &$ip) {
```
cedric-anne, line 384:
```suggestion
                        foreach ($device_ips as $device_ip) {
```

inc/networkinventory.class.php
```diff
@@ -367,11 +367,33 @@ public function run($jobstate)
...
+                            if ($pfIPRange->getIp2long($ip) <= $pfIPRange->getIp2long($pfIPRange->fields['ip_end']) && $pfIPRange->getIp2long($pfIPRange->fields['ip_start']) <= $pfIPRange->getIp2long($ip)) {
+                                // in range, assign this IP as device IP
+                                break;
+                            }
+                        }
```
cedric-anne, line 389:
```suggestion
                        foreach ($device_ips as $device_ip) {
                            if ($pfIPRange->getIp2long($device_ip) <= $pfIPRange->getIp2long($pfIPRange->fields['ip_end']) && $pfIPRange->getIp2long($pfIPRange->fields['ip_start']) <= $pfIPRange->getIp2long($device_ip)) {
                                // in range, assign this device IP
                                $id = $device_ip;
                                break;
                            }
                        }
```

inc/networkinventory.class.php
```diff
@@ -367,11 +367,33 @@ public function run($jobstate)
...
+                        break;
+                        
+                }
+            }
+        }
+        if(!isset($ip)) $ip = current($ips);
```
cedric-anne, line 395:
```suggestion
```

cedric-anne (changes requested):

inc/networkinventory.class.php
stonebuzz: Why remove this fallback ? (if it's not a PluginGlpiinventoryIPRange or if the IPs don't match the Range)

cedric-anne: Value is assigned outside the lopp in the first suggestion.
image

I think is is easier to understand and it prevents the usage of a reference in the loop.

cedric-anne (approved):

eduardomozart:
Hello guys,
I did some changes on the script to handle cenarios where there's some sort of VIP (virtual IP, used for redundancy on network devices).
On my infrastructure, we have an Aruba Instant Access Points (IAP) cluster (a virtual controller that handle all APs on our site), as you can see below, both devices contains "192.168.0.6", which is a VIP that I use to manage the IAP cluster (only one device assume this IP at the time, and if this AP is down, other devices assume this VIP, so the cluster management is always available), but each AP has it's own AP:
image
image
So this virtual IP (192.168.0.6) is on the same range than my task (192.168.0.0/24), but it will be used only if there's no other unique IP available on the same network range.

trasher:
This would need tests

trasher:
Please fix CS

eduardomozart:
> This would need tests

I'll see what I can do.

> Please fix CS

Which line(s)?

trasher:
> > Please fix CS
> Which line(s)?

See GH actions output: https://github.com/glpi-project/glpi-inventory-plugin/actions/runs/9891523512/job/28031989775?pr=524

eduardomozart:
Hello,
I added the Unit Tests as requested by trasher but as cedric already approved those changes and the changes I made after his approval was minimal (only continued loop if IP wasn't unique instead of breaking it), I believe it should work as expected.
The changes I made after cedric approval may not be needed (at least, for Aruba IAP cluster Virtual IPs) if #536 be investigated further, but maybe other systems of redundancy can also assign an VIP for more than one device, so this kind of check maybe useful to keep.

stonebuzz (approved):

trasher (approved):

PR #526, opened 2024-07-11
Fix translation of strings

eduardomozart:
1. It isn't being translated on PT-BR language.

This fix also requires add the missing "``%s inventoried``" string to the ``.po`` files. E.g. to PT-BR:

```
#: inc/collect_registry_content.class.php:128
#: inc/collect_registry_content.class.php:138
#: inc/collect_registry_content.class.php:148
#: inc/collect_registry_content.class.php:158
msgid "%s inventoried"
msgstr "%s inventariados(as)"
```

2. I also noticed that when creating the Dashboard, the Dashboard title isn't translatable:

https://github.com/glpi-project/glpi-inventory-plugin/blob/main/install/update.php#L1106

But native GLPI dashboard are:

image

I couldn't figure it out how to translate it.

3. I believe the word "Time slot" should also be available on plural form.
4. I also noticed that when changing the GLPI UI language to another one (e.g. Portuguese to English), the GLPI Inventory plugin still shows the menus on Portuguese. I'm not sure, but I believe it maybe related to menu.class.php file and a similar issue as the item number 2 of this list. It seems to always use the language that the user was using when installed the plug-in.
5. I found a non-translated English string on ``taskjobs.js`` file that needs to be added to the translation files. E.g. for PT-BR:

```
#: js/taskjobs.js:715
msgid "The preparation of the task did not return any targets"
msgstr "A preparação da tarefa não retornou nenhum alvo"
```

### DIFF
--- a/inc/timeslotentry.class.php
+++ b/inc/timeslotentry.class.php
@@ -216,7 +216,7 @@ class PluginGlpiinventoryTimeslotEntry extends CommonDBTM
             echo "</td>";
             echo "<td colspan='2'>";
             if ($canedit) {
-                echo "<input type='submit' class='submit' name='purge-" . $dbentry['id'] . "' value='".__('Delete')."' />";
+                echo "<input type='submit' class='submit' name='purge-" . $dbentry['id'] . "' value='" . __('Delete') . "' />";
             }
             echo "</td>";
             echo "</tr>";
--- a/inc/timeslotentry.class.php
+++ b/inc/timeslotentry.class.php
@@ -216,7 +216,7 @@ class PluginGlpiinventoryTimeslotEntry extends CommonDBTM
             echo "</td>";
             echo "<td colspan='2'>";
             if ($canedit) {
-                echo "<input type='submit' class='submit' name='purge-" . $dbentry['id'] . "' value='delete' />";
+                echo "<input type='submit' class='submit' name='purge-" . $dbentry['id'] . "' value='".__('Delete')."' />";
             }
             echo "</td>";
             echo "</tr>";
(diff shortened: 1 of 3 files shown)


js/taskjobs.js
```diff
@@ -712,7 +712,7 @@ taskjobs.update_logs = function (data) {
          var targets_cpt = 0;
 
          if(job_v.targets.length === 0) {
-            $(targets_selector).html('The preparation of the task did not return any targets');
+            $(targets_selector).html(__('The preparation of the task did not return any targets', 'glpiinventory'));
```
stonebuzz, line 715: From `js` file call to `__('my string', 'context')` is not possible as this is a PHP function

stonebuzz (changes requested):

js/taskjobs.js
eduardomozart: Hello @stonebuzz, it’s now possible since GLPI 9.5.0. I tested it and it’s working as expected. All PHP functions was backported to JavaScript: https://glpi-developer-documentation.readthedocs.io/en/master/devapi/translations.html

stonebuzz: you're right (it's been so long since I've done gettext in js :smile:)

you can ignore this comment

inc/timeslotentry.class.php
```diff
@@ -216,7 +216,7 @@ public function formDeleteEntry($timeslots_id)
             echo "</td>";
             echo "<td colspan='2'>";
             if ($canedit) {
-                echo "<input type='submit' class='submit' name='purge-" . $dbentry['id'] . "' value='delete' />";
+                echo "<input type='submit' class='submit' name='purge-" . $dbentry['id'] . "' value='".__('Delete')."' />";
```
trasher, line 219:
```suggestion
                echo "<input type='submit' class='submit' name='purge-" . $dbentry['id'] . "' value='" . __('Delete') . "' />";
```

PR #527, opened 2024-07-11
Update es_ES.po

jjcamarasa:
After use by me and a colleague at work, we found some strange words and missing translations in package deployment.
Thanks for all, the glpi soft and the agent.

Writed missing translations.
Changed dumb text like in original Choose -> Eleja, using Elegir (because Eleja is not a word in Spanish) or bad translations like suceso for succesful that is correcto in spanish.

### DIFF
--- a/locales/es_ES.po
+++ b/locales/es_ES.po
@@ -89,11 +89,11 @@ msgstr "Añadir a grupo estático"
 
 #: inc/deployuserinteraction.class.php:104
 msgid "After actions"
-msgstr "Después de acciones"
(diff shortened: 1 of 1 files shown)


jjcamarasa:
> Hi @jjcamarasa
> 
> plugin translations are managed on the Transifex service.
> 
> https://app.transifex.com/teclib/glpi-plugin-glpiinventory/dashboard/
> 
> It is on this platform that you can modify / complete the translations.

OK, I'm going there. Thanks!

PR #529, opened 2024-07-11
Fix NetDiscovery state number display

eduardomozart:
Pagination works on NetInventory state page, but not on NetDiscovery one. This fix it.
Affected URI: https://glpi.example.com/plugins/glpiinventory/front/statediscovery.php

### DIFF
--- a/inc/statediscovery.class.php
+++ b/inc/statediscovery.class.php
@@ -165,7 +165,7 @@ class PluginGlpiinventoryStateDiscovery extends CommonDBTM
             'ORDER' => 'uniqid DESC'
         ]);
         $result = $iterator->next();
-        $number = $result['cpt'];
+        $number = count($iterator);
 
         // Display the pager
         Html::printPager($start, $number, Plugin::getWebDir('glpiinventory') . "/front/statediscovery.php", '');
--- a/inc/statediscovery.class.php
+++ b/inc/statediscovery.class.php
@@ -164,7 +164,6 @@ class PluginGlpiinventoryStateDiscovery extends CommonDBTM
             'GROUP' => 'uniqid',
             'ORDER' => 'uniqid DESC'
         ]);
-        $result = $iterator->next();
         $number = count($iterator);
 
         // Display the pager


stonebuzz (approved):

inc/statediscovery.class.php
```diff
@@ -165,7 +165,7 @@ public function display($options = [])
             'ORDER' => 'uniqid DESC'
         ]);
         $result = $iterator->next();
-        $number = $result['cpt'];
+        $number = count($iterator);
```
trasher, line 168: I'm surprised this works. Iterator is a "COUNT", tehrefore it always return only 1 result. That seems no correct.

trasher: Maybe using `$result = $iterator->current();` would solve the issue

eduardomozart: Hello @trasher,
I'm not sure, as the $number variable is necessary to render the table on L171.
I applied the same fix currently on production in https://github.com/glpi-project/glpi-inventory-plugin/blob/main/inc/stateinventory.class.php#L106

trasher: I have no idea what is really expected here. My concern is the usage of the dbiterator: either it's not used correctly, or the query should not be a count, no idea). 

Consider the following:
```
$iterator = $DB->request([
    'FROM' => 'glpi_computers'
]);
var_dump(count($iterator)); ===> 3

$iterator = $DB->request([
    'COUNT' => 'cpt',
    'FROM' => 'glpi_computers'
]);
var_dump(count($iterator)); ===> 1
$row = $iterator->current();
var_dump($row['cpt']); ==> 3
```

Do you see the difference?

But even worse, if the result of the count is `0`, the iterator will always return one row:
```
$iterator = $DB->request([
    'FROM' => 'glpi_appliances'
]);
var_dump(count($iterator)); ===> 0

$iterator = $DB->request([
    'COUNT' => 'cpt',
    'FROM' => 'glpi_appliances'
]);
var_dump(count($iterator)); ===> 1
$row = $iterator->current();
var_dump($row['cpt']); ===> 0
```

So proposed fix is not correct; `$number` will always be `1` whatever the SQL query returns.

trasher (changes requested):
See comments

inc/statediscovery.class.php
trasher: OK, my bad; the group gives several rows with count, and it seems we indeed must rely oon the number of lines the query returns

trasher (approved):

PR #532, opened 2024-07-17
fix(collect): display columns in search

Rom1-B:
ref !33786

The columns could not be added in the computer search view parameters (button: select default items to show), although they could be searched.

_It seems to have been deactivated 8 years ago, but I haven't found the reason?_

### DIFF
--- a/inc/collect.class.php
+++ b/inc/collect.class.php
@@ -40,33 +40,33 @@ if (!defined('GLPI_ROOT')) {
  */
 class PluginGlpiinventoryCollect extends CommonDBTM
 {
-   /**
-    * The right name for this class
-    *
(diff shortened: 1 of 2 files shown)


stonebuzz (approved):
seems OK to me

trasher:
> _It seems to have been deactivated 8 years ago, but I haven't found the reason?_

No idea neither. Seems it has been added on https://github.com/glpi-project/glpi-inventory-plugin/commit/701aad61f94f747541fe6861128d7100001324a3; maybe @orthagh remembers why it's been explicitly deactivated?

orthagh (approved):
Questions about decisions from 8 years ago, no way I remember

PR #540, opened 2024-09-03
Fix(Core): Optimize dynamic group display

stonebuzz:
When we display or update a dynamic group

The plugin tries to calculate all the computers (linked to the search) to keep the data in the database (like a cache).

It doesn't need to, because displaying the dynamic group also shows the computers involved

and `getTargetsForGroup`, executes the search twice ...

(so three times in total when displaying a group)


trasher (approved):
I cannot easily test, and you certainly know much well that part than I do

PR #544, opened 2024-09-06
1.3.6

stonebuzz:
### Fixed

- The columns could not be added in the computer search view (collect).
- Fix `invalid item` for obsolete agents.
- Fix upload directory set in configuration.
- Fix `cancel` count from job executions.
- Fix the management of the `include old jobs` parameter.
- Use Device IP in IP range set on task for Device NetInventory
- Fix pagination for `NetInventory` state page

### Feat

- Display all computers details for dynamics group
- Count agent handle by `Taskscheduler`

### Changes

- Encrypt credentials
- Displays ```Tasks / Groups``` tab only on computers linked to an agent

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
-## [1.3.6] - 2024-09-06
+## [1.4.0] - 2024-09-06
 
 ### Fixed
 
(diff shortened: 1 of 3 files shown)


Rom1-B (changes requested):
As there are also features and changes, shouldn't the version be **1.4.0**?

Rom1-B (commented):
OK, but perhaps it would be more prudent to recreate the branch with the new number?

trasher (approved):

PR #575, opened 2024-12-02
Remove translation of technical term

### LINKED ISSUE #574: "moniker" definition in WMI Collect related code
### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

### Is there an existing issue for this?

- [x] I have searched the existing issues

### GLPI Version

10.0.17

### Plugin version

1.4.0

### Bug description

"moniker" is a technical term which should not be translated IMHO. For example it is translated as "surnom" in french, "Name" in german (yes first letter is capitalized) or "alias" in spanish.

Also it could be nice to show an example as indeed glpi-agent expects a specific format. For example the default comportment if the value is empty is to use `winmgmts://./root/cimv2`. If someone wants to use [Root\CIMV2\Security\MicrosoftTpm as Namespace for Win32_Tpm](https://learn.microsoft.com/en-us/windows/win32/secprov/win32-tpm), he have to define moniker with `winmgmts://./Root/CIMV2/Security/MicrosoftTpm`.

Related ticket: `!35436`

### Relevant log output

```shell

```

### Page URL

_No response_

### Steps To reproduce

_No response_

### Your GLPI setup information

_No response_

### Anything else?

_No response_

trasher:
closes #574

### DIFF
--- a/inc/collect_wmi.class.php
+++ b/inc/collect_wmi.class.php
@@ -57,7 +57,7 @@ class PluginGlpiinventoryCollect_Wmi extends PluginGlpiinventoryCollectCommon
     {
         return [
          __("Name"),
-         __("Moniker", "glpiinventory"),
+         "Moniker",
          __("Class", "glpiinventory"),
          __("Properties", "glpiinventory"),
          __("Action")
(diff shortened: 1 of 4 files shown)


orthagh (approved):

g-bougard (approved):
Thank you

PR #580, opened 2024-12-17
Fix collect registries content limit

Lainow:

## Description

- It fixes !35638

A collection task only receives values of 255 characters, whereas the registry key can contain more. This causes errors for some customers


trasher:
Did you test if update goes well? For this plugin, it always has been a nightmare :D

trasher (approved):

stonebuzz (approved):

PR #581, opened 2024-12-18
fix: PHP error - allowed memory size

Rom1-B:

## Description

- It fixes !35468

Prevent this type of error message :
```
glpiphplog.CRITICAL:   *** PHP Error (1): Allowed memory size of 268435456 bytes exhausted (tried to allocate 48238592 bytes) in /data/glpi/plugins/glpiinventory/inc/task.class.php at line 1347
```

The `logIfExtradebug()` function is used to log data when the "extra-debug" option is enabled. However, when the data size is very large and this option is disabled, it triggers a memory exhaustion error when the function is called, even though the function itself does nothing.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,8 +8,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ## [Unreleased]
 
 ### Fixed
+
 - Ensure that the `Taskjob` identifier is used to extract the device IP.
 - Isolate dynamic group criteria to prevent their global reapplication in GLPI.
+- Fixes memory exhaustion when "extra-debug" is disabled
 
 ## [1.4.0] - 2024-09-06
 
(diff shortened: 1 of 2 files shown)


trasher (approved):

Rom1-B:
I've set it to draft because I'd like confirmation from the client to see if this is sufficient or if we need to go further.

trasher:
Could you please rebase to solve conflicts?

Rom1-B:
I don't see any conflicts???

image

trasher:
I have on my side... :/
image

trasher:
I guess this is your last merge commit that cause issue.

Rom1-B:
Do you still see conflict?

trasher:
> Do you still see conflict?

Nope, it's OK; thanks :)

PR #587, opened 2025-01-02
Fix SNMP communication error (wrong tasks id gived)

Lainow:

## Description

- It fixes !35754

Previously, the value used when calling the `getDeviceIPOfTaskID` method was the taskjob id, but the requested value was the task id,

### DIFF
--- a/inc/networkinventory.class.php
+++ b/inc/networkinventory.class.php
@@ -367,7 +367,10 @@ class PluginGlpiinventoryNetworkinventory extends PluginGlpiinventoryCommunicati
         $current = $jobstate;
         $agent->getFromDB($current->fields['agents_id']);
 
-        $ip = $this->getDeviceIPOfTaskID($jobstate->fields['itemtype'], $jobstate->fields['items_id'], $jobstate->fields['plugin_glpiinventory_taskjobs_id']);
+        $taskjob = new PluginGlpiinventoryTaskjob();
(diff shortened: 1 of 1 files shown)


g-bougard:
In my own tests, I found the code was working as far as you create one job by task: taskjob id remains aligned with task id.

Then if you create a task without any job or if you purge a job to recreate a new one on a task, then ids are no more aligned. New tasks or task with purged job will only provide the expected device ip with this patch applied.

inc/networkinventory.class.php
```diff
@@ -367,7 +367,10 @@ public function run($jobstate)
         $current = $jobstate;
         $agent->getFromDB($current->fields['agents_id']);
 
-        $ip = $this->getDeviceIPOfTaskID($jobstate->fields['itemtype'], $jobstate->fields['items_id'], $jobstate->fields['plugin_glpiinventory_taskjobs_id']);
+        $taskjob = new PluginGlpiinventoryTaskjob();
+        $taskjob->getFromDB($jobstate->fields['plugin_glpiinventory_taskjobs_id']);
+
```
stonebuzz, line 372:
```suggestion
$current = $jobstate;
```

stonebuzz: `$current` is used later we need to update it to make sure it contains the good `jobstate`

stonebuzz (approved):

Rom1-B (approved):

PR #599, opened 2025-01-22
Translate export columns of task results

eduardomozart:
When clicking on the Export results of tasks button, the column titles of the exported file are in English.

## Description

### DIFF
--- a/inc/taskview.class.php
+++ b/inc/taskview.class.php
@@ -594,15 +594,15 @@ class PluginGlpiinventoryTaskView extends PluginGlpiinventoryCommonView
         }
 
         // cols titles
-        echo "Task_name" . SEP;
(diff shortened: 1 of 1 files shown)


eduardomozart:
Decided to close this PR since other informations are already shown in English, so it make sense to keep the column in English, too. If handling translation of ``last_log`` entry on CSV file, the PHP function ``normalizer_normalize($exec['last_log'], Normalizer::FORM_C)`` should be implemented since characters with accents (é) aren't being rendered as expected on Excel because most TTF fonts are Unicode instead of UTF-8, e.g. in UTF-8, the character "é" can be represented by "e + '", while Unicode represents it as an entity "é". As most fonts doesn't support UTF-8, they will print "e" followed by an square, as can be seen on the print below:
image

PR #600, opened 2025-01-22
Fix translation of crontask descriptions

eduardomozart:
GLPI is multilanguage so translating the comment it uses the language of the user when the plug-in has been installed.

## Description

### DIFF
--- a/inc/agentwakeup.class.php
+++ b/inc/agentwakeup.class.php
@@ -48,6 +48,19 @@ class PluginGlpiinventoryAgentWakeup extends CommonDBTM
     public static $rightname = 'plugin_glpiinventory_taskjob';
 
 
+    /**
+     * Get cron task's description
+     *
+     * @return array
+     */
+    public static function cronInfo(): array
+    {
+        return [
+            'description' => __('Wake agents ups', 'glpiinventory')
+        ];
+    }
+
+
    /**
     * Get name of this type by language of the user connected
     *
--- a/inc/agentwakeup.class.php
+++ b/inc/agentwakeup.class.php
@@ -56,7 +56,7 @@ class PluginGlpiinventoryAgentWakeup extends CommonDBTM
     public static function cronInfo(): array
     {
         return [
-            'description' => __('Wake agents ups', 'glpiinventory')
+            'description' => __('Wake agents up', 'glpiinventory')
         ];
     }
 
--- a/install/update.php
+++ b/install/update.php
@@ -888,26 +888,15 @@ function pluginGlpiinventoryUpdate($current_version, $migrationname = 'Migration
     if ($crontask->getFromDBbyName('PluginGlpiinventoryTaskjob', 'updatedynamictasks')) {
         $crontask->delete($crontask->fields);
     }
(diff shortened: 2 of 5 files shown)


cconard96 (commented):
Instead of using a comment at all, these classes should define a `cronInfo` function that returns the translated description IMO.
`cronInfo` takes a string that is the task name and returns an array like:
```
return ['description' => __('Description of the requested task')];
```

eduardomozart:
Hello @cconard96, thank you for your suggestion, I didn't know that function exists. I updated the PR to use them and also fixes some descriptions that wasn't being loaded because the crontask has been registered with lowercase characters.

inc/agentwakeup.class.php
```diff
@@ -48,6 +48,19 @@ class PluginGlpiinventoryAgentWakeup extends CommonDBTM
...
 
+    /**
+     * Get cron task's description
+     *
+     * @return array
+     */
+    public static function cronInfo(): array
+    {
+        return [
+            'description' => __('Wake agents ups', 'glpiinventory')
```
stonebuzz, line 59: I have a doubt about this translation 

@cconard96 can you validate (or not)?

install/update.php
```diff
@@ -888,26 +888,15 @@ function pluginGlpiinventoryUpdate($current_version, $migrationname = 'Migration
...
-            86400,
-            ['mode' => 2, 'allowmode' => 3, 'logs_lifetime' => 30,
-                               'hourmin' => 22, 'hourmax' => 6,
-            'comment' => Toolbox::addslashes_deep(__(
-                'Delete agents that have not contacted the server since "xxx" days.',
-                'glpiinventory'
-            ))]
-        );
+    if ($crontask->getFromDBbyName('PluginGlpiinventoryAgent', 'cleanoldagents')) {
+        $crontask->delete($crontask->fields);
```
stonebuzz, line 892: Why this change ?

stonebuzz (changes requested):

install/update.php
eduardomozart: It's a deleted crontask as can be seen in https://github.com/glpi-project/glpi-inventory-plugin/blob/bc7d2cddeaf7641e511a5a2b3a881c54d6e5a0f4/install/update.native.php#L700, I believe it becomes native in GLPI 10 since the agent management was migrated to the core.

inc/agentwakeup.class.php
eduardomozart: A way to validate it is simply apply it, go to "Crontasks" (Automatic actions) on GLPI and see if the "Description" field is filled. This function is already used on GLPI-Inventory plug-in itself: https://github.com/glpi-project/glpi-inventory-plugin/blob/bc7d2cddeaf7641e511a5a2b3a881c54d6e5a0f4/inc/task.class.php#L756-L762

cconard96, line 59:
```suggestion
            'description' => __('Wake agents up', 'glpiinventory')
```

cconard96: I think he was only asking me to validate the description sounded correct in English

stonebuzz (approved):

PR #601, opened 2025-01-22
Translate error messages for end user

eduardomozart:
The plug-in stores the error message based on the current user language. This PR stores the message on English and translates the message on the client side, so on GLPI instances with multilanguage user, the task log will show the error message based on the current user language instead of the language of the user that the error message has been generated.

Before this PR, an Portuguese language user was seeing error messages on English:

image

After this PR, a Portuguese Language user can see error messages in their native language independent of the language of the user that the error has been generated:

image

Edit: The only known issue I can think of this PR can cause is that those strings would be removed from the POT file, so I believe it would need a way to keep them there or do a dummy call/register of them into JS/PHP code (maybe a fake/unused property of the same object that I modified into this PR) so they would be gathered by Transiflex bot and wouldn't be missing.

## Description

### DIFF
--- a/inc/task.class.php
+++ b/inc/task.class.php
@@ -479,7 +479,7 @@ class PluginGlpiinventoryTask extends PluginGlpiinventoryTaskView
             if ($result['task']['is_active'] == 0) {
                 $msg = "The task has been deactivated after preparation of this job.";
                 $translatable_msg = __("The task has been deactivated after preparation of this job.", 'glpiinventory');
-                
+
                 $jobstates_to_cancel[$jobstate->fields['id']] = [
                 'jobstate' => $jobstate,
                 'reason'   => $msg
--- a/inc/task.class.php
+++ b/inc/task.class.php
@@ -464,9 +464,12 @@ class PluginGlpiinventoryTask extends PluginGlpiinventoryTaskView
                 $result['run']['state'] == PluginGlpiinventoryTaskjobstate::SERVER_HAS_SENT_DATA
                  or $result['run']['state'] == PluginGlpiinventoryTaskjobstate::AGENT_HAS_SENT_DATA
             ) {
+                $msg = "The agent is requesting a configuration that has already been sent to him by the server. It is more likely that the agent is subject to a critical error.";
+                $translatable_msg = __("The agent is requesting a configuration that has already been sent to him by the server. It is more likely that the agent is subject to a critical error.", 'glpiinventory');
+
                 $jobstates_to_cancel[$jobstate->fields['id']] = [
                 'jobstate' => $jobstate,
-                'reason'   => "The agent is requesting a configuration that has already been sent to him by the server. It is more likely that the agent is subject to a critical error.",
+                'reason'   => $msg,
                 'code'     => $jobstate::IN_ERROR
                 ];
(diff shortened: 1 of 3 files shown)


trasher (changes requested):
As far as I can see, strings will never be translated at all, because sources won't be extracted to POT file.

stonebuzz:
I think so, thanks to this

image

Now the string is translated on demand (depending on the user's language)

trasher:
@stonebuzz nope. Strings have to exist in source code, and be "marked" as translatable for extraction to work.

stonebuzz:
> @stonebuzz nope. Strings have to exist in source code, and be "marked" as translatable for extraction to work.
> 
> So this code cans indeed give a translated value, my comment concerns source strings extraction from source code.

I see, you're right

trasher:
> We should keep the call to gettext (__(‘string to translate’, ‘context’) so that the CI can extract the string and translate it in Transifex.
> 
> ex :
> 
> ```
> if (
>     $result['run']['state'] == PluginGlpiinventoryTaskjobstate::SERVER_HAS_SENT_DATA
>         or $result['run']['state'] == PluginGlpiinventoryTaskjobstate::AGENT_HAS_SENT_DATA
> ) {
>     $msg = "The agent is requesting a configuration that has already been sent to him by the server. It is more likely that the agent is subject to a critical error.";
>     $translatable_msg = __($msg, 'glpiinventory');
> 
>     $jobstates_to_cancel[$jobstate->fields['id']] = [
>         'jobstate' => $jobstate,
>         'reason'   => $msg
>         'code'     => $jobstate::IN_ERROR
>     ];
>     continue;
> }
> ```

Nope, this is still wrong :/

String to translate must be in gettext function (gettext extraction will not execute any php code):
```
    $msg = "The agent is requesting a configuration that has already been sent to him by the server. It is more likely that the agent is subject to a critical error.";
    $translatable_msg = __("The agent is requesting a configuration that has already been sent to him by the server. It is more likely that the agent is subject to a critical error.", 'glpiinventory');
```

eduardomozart:
I updated this PR following the instructions from @stonebuzz and @trasher.

trasher (approved):

stonebuzz (approved):

trasher:
Tests are failing... :/ 
It's surprising; maybe test itself is incorrect - I do not have time to check.

eduardomozart:
Hello @trasher, now the tests works as expected.

trasher:
I do not understand the error, and I do not understand the fix neither. Nor why this is breaking now - seems unrelated.

As far a we can see, it can have several causes; I'd prefer to be sure understanding the fail.

stonebuzz:
Oo => https://github.com/glpi-project/glpi-inventory-plugin/pull/607

no error XD

stonebuzz:
@eduardomozart 

can you revert last commit about test ? to try

eduardomozart:
I believe there should be an error on my branch, I'm unable to commit on it using GitHub Desktop, only through GitHub WebUI. I'll try to revert my last commit.

stonebuzz:
:smile: 

maybe a problem with Github? 
I doubt the problem is with your branch

PR #613, opened 2025-01-27
Improve translation

eduardomozart:
This is a revision of PR#601. When translating on client-side, the translation may fail for some items (e.g. "updatetheitem" strings) because it creates a string on-the-fly (e.g. "Update the item Computer", "Update the item Network Device"). Also, when translating on client-side, it creates some artifacts when calling the ``$Class->getLink()`` call because the link are escaped and isn't rendered as HTML on client-side.

## Description

### DIFF
--- a/inc/taskjoblog.class.php
+++ b/inc/taskjoblog.class.php
@@ -347,6 +347,8 @@ class PluginGlpiinventoryTaskjoblog extends CommonDBTM
     public static function convertComment($comment)
     {
         $matches = [];
+       // Attempt to translate fixed strings (e.g. error messages, see PR#601)
+        $comment = __($comment, 'glpiinventory');
        // Search for replace [[itemtype::items_id]] by link
         preg_match_all("/\[\[(.*)\:\:(.*)\]\]/", $comment, $matches);
         foreach ($matches[0] as $num => $commentvalue) {
@@ -354,7 +356,7 @@ class PluginGlpiinventoryTaskjoblog extends CommonDBTM
             if ($classname != '' && class_exists($classname)) {
                 $Class = new $classname();
                 $Class->getFromDB($matches[2][$num]);
-                $comment = str_replace($commentvalue, $Class->getLink(), $comment);
+                $comment = $Class->getTypeName() + " " + str_replace($commentvalue, $Class->getLink(), $comment);
             }
         }
         if (strstr($comment, "==")) {


inc/taskjoblog.class.php
```diff
@@ -346,34 +346,54 @@ public function getDivState($state, $type = 'div')
...
-                $comment = str_replace($commentvalue, $Class->getLink(), $comment);
+                $Class->getFromDB($matches[3][$num]);
+                $comment = str_replace($commentvalue, $Class->getTypeName(1) . " " . $Class->getLink(), $comment);
+            } else {
+                $comment = str_replace($commentvalue, $matches[1][$num], $comment);
             }
         }
+
         if (strstr($comment, "==")) {
             preg_match_all("/==([\w\d]+)==/", $comment, $matches);
```
stonebuzz, line 379: All occurence of `==devicesfound==` (and other) have been drop from this PR, this might not work, right?

stonebuzz:
Thanks for your PR and for the hard work you've put into it. I've had a good look at the changes, and I have to admit that I'm a bit wary of incorporating such a major modification, especially in such a sensitive part of the code.

My main concern is the maintainability and stability of this critical section. With a large volume of modifications, it becomes more difficult to anticipate the impact and ensure that no side-effects emerge.

Honestly, I don't have the time/resources to test all the use cases impacted by these changes.

The risk of bugs/improvements is far too great to consider merging it for now.

I know you spend quite a bit of time on the plugin, and it's great to see that despite its complexity, you’re sticking with it. But honestly, this plugin is destined to disappear in favor of a completely rewritten version, and we have already identified these kinds of "gaps" or "oddities" that will need to be drastically improved.

inc/taskjoblog.class.php
eduardomozart: Static text still uses the old ``==string==`` syntax, e.g. "==actioncancelled==" returns "Action cancelled by user" (there's no dynamic content, e.g. numbers into it).

eduardomozart:
I see, no problem. This commit only does the following to ``convertComment`` function:

1. Replace all instances of "[[]]" to place holders (first foreach), which allows them to be gathered by gettext, e.g. "[[2]] devices found" becomes "%1$2 devices found". Every instance of "[[]]" will receive a number (%1$s, %2$s and go on).
2. Now it's static! It means we can translate them using gettext. It also means the user can move the numbering through the PO file if the user language requires it, e.g. "Dispositivos encontrados %1$s" in Portuguese).
3. String is translated, let's restore the placeholder value with it's respective value (second foreach). Into this example, it would return "Dispositivos encontrados [[2]]".

The rest of the logic remains the same, only the second argument of ``[[]]`` becomes optional, so it supports ``[[ItemType::ItemID]]`` and ``[[<String>]]`` syntax. If it is ``[[ItemType::ItemID]]``, it will translate as "Dispositivo de Rede <Name with Hyperlink to Asset>" as it is today (no changes), but if it uses the ``[[<String>]]``, it will simply remove the brackets from <String>, so "Dispositivos encontrados [[2]]" would become "Dispositivos encontrados 2" into the final translation.

If this change is too cubersome, please let me know that I remove those changes and keep the code as it is, only mantaining the added static string which hasn't been added to the translation before.

stonebuzz:
At the risk of repeating myself, the likelihood of bugs or unforeseen improvements is too high to consider a merge at this stage.

That being said, this enhancement, while highly relevant in substance, is an improvement I would like to see integrated into the next version of the plugin.

In the meantime, feel free to keep this enhancement as a branch or patch in your repository.

For now, I prefer to avoid any major changes to this plugin.

eduardomozart:
No problem @stonebuzz. I don't think to create a private fork/branch for myself, I think that if my commits can't benefit the community, than it's better not to implement it.
I updated the commits and now this is a simplified version of the original function, only with additional translated strings.

PR #616, opened 2025-01-28
Replace deploy icons by sprites

eduardomozart:

Before this PR:

SCR-20250128-mqrl

After this PR:

image

## Description


stonebuzz (commented):
Can you delete related pics if not used anymore ? Best regards

eduardomozart:
> Can you delete related pics if not used anymore ? Best regards

Done =D

css/deploy.css
```diff
@@ -138,10 +142,11 @@ a.edit {
 }
 /* drag and drop element */
 .drag {
-   background-image: url("../pics/drag.png");
+   /* background-image: url("../pics/drag.png"); */
```
stonebuzz, line 145: can be removed instead of commented

stonebuzz (changes requested):

inc/deployaction.class.php
```diff
@@ -439,7 +439,7 @@ public function displayAjaxValues($config, $request_data, $rand, $mode)
                      'value' => (isset($config_data['logLineLimit'])) ? $config_data['logLineLimit'] : 10
                     ];
             Dropdown::showNumber('logLineLimit', $options);
-            echo "&nbsp;<span class='red'><i>";
+            echo "&nbsp;<i class='ti ti-exclamation-circle' style='color: #FFA500; vertical-align: middle;'></i><span class='red'><i>";
```
trasher, line 442: Please avoid using inline styles; use CSS proper classes instead.

css/deploy.css
eduardomozart: Done

eduardomozart:
Made the requested changes

stonebuzz (approved):

css/deploy.css
```diff
@@ -29,6 +29,10 @@
  * ---------------------------------------------------------------------
  */
 
+.select2-container--default .select2-results__group {
+   font-weight: 600;
```
trasher, line 33: Won't that affect core and other plugins select2 dropdowns?

eduardomozart: Not a big fan about it too, better would be adding a !important flag on GLPI core, since the Deploy action includes ``ext-js`` which overrides this default GLPI CSS and not apply it. Actually this is the same value of GLPI core, I'll open a PR on GLPI Core for it.

stonebuzz (approved):

trasher (approved):

PR #621, opened 2025-01-29
Update translation of Deploy Groups

eduardomozart:

Also add icon to the Deploy Group:

image

## Description

### DIFF
--- a/inc/deploygroup.class.php
+++ b/inc/deploygroup.class.php
@@ -98,7 +98,7 @@ class PluginGlpiinventoryDeployGroup extends CommonDBTM
     */
     public static function getTypeName($nb = 0)
     {
-        return __('Inventory group', 'glpiinventory');
+        return _n('Inventory group', 'Inventory groups', $nb, 'glpiinventory');
     }
 
 
--- a/inc/deploygroup.class.php
+++ b/inc/deploygroup.class.php
@@ -114,7 +114,11 @@ class PluginGlpiinventoryDeployGroup extends CommonDBTM
         $this->addDefaultFormTab($ong);
 
         $count = self::getMatchingItemsCount("PluginGlpiinventoryTaskjob");
-        $ong[$this->getType() . '$task'] = self::createTabEntry(_n('Associated task', 'Associated tasks', Session::getPluralNumber(), 'glpiinventory'), $count);
+        if ($_SESSION['glpishow_count_on_tabs']) {
+            $tabs[2] = self::createTabEntry(_n('Associated task', 'Associated tasks', Session::getPluralNumber(), 'glpiinventory'), $count);
+        } else {
+            $tabs[2] = _n('Associated task', 'Associated tasks', Session::getPluralNumber());
+        }
 
         $this->addStandardTab('Log', $ong, $options);
         return $ong;
--- a/inc/deploygroup.class.php
+++ b/inc/deploygroup.class.php
@@ -114,7 +114,7 @@ class PluginGlpiinventoryDeployGroup extends CommonDBTM
         $this->addDefaultFormTab($ong);
 
         $count = self::getMatchingItemsCount("PluginGlpiinventoryTaskjob");
-        $ong[$this->getType() . '$task'] = self::createTabEntry(_n('Associated task', 'Associated tasks', $count), $count);
+        $ong[$this->getType() . '$task'] = self::createTabEntry(__('Associated tasks', 'glpiinventory'), $count);
 
         $this->addStandardTab('Log', $ong, $options);
         return $ong;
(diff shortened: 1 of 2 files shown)


inc/deploygroup.class.php
```diff
@@ -114,7 +114,7 @@ public function defineTabs($options = [])
         $this->addDefaultFormTab($ong);
 
         $count = self::getMatchingItemsCount("PluginGlpiinventoryTaskjob");
-        $ong[$this->getType() . '$task'] = self::createTabEntry(_n('Associated task', 'Associated tasks', $count), $count);
+        $ong[$this->getType() . '$task'] = self::createTabEntry(__('Associated tasks', 'glpiinventory'), $count);
```
trasher, line 117: I do not agree with this change. singular/plural form exists elsewhere, it's better to keep only one occurrence for translation.

inc/deploygroup.class.php
```diff
@@ -819,4 +819,9 @@ public function showForComputer($computers_id)
         }
         echo "</table>";
     }
+
+
+    public static function getIcon() {
```
trasher, line 824: Any idea what the result is, a screenshot would have been welcome (we cannot take time to test every small change proposal).
Also, the method breaks the CI.

trasher (changes requested):

inc/deploygroup.class.php
eduardomozart: Updated this PR to include screenshot.

inc/deploygroup.class.php
```diff
@@ -114,7 +114,7 @@ public function defineTabs($options = [])
         $this->addDefaultFormTab($ong);
 
         $count = self::getMatchingItemsCount("PluginGlpiinventoryTaskjob");
-        $ong[$this->getType() . '$task'] = self::createTabEntry(_n('Associated task', 'Associated tasks', $count), $count);
+        $ong[$this->getType() . '$task'] = self::createTabEntry(_n('Associated task', 'Associated tasks', Session::getPluralNumber(), 'glpiinventory'), $count);
```
trasher, line 117: You've reverted your change, but you've also updated the other translation to always use `Session::getPluralNumber` - Therefore singular is now no longer used.

Also, not directly related to your PR, but this method does not respect `glpi_count_on_tabs` parameter, I guess it should.

eduardomozart: This follow the standard of all other tabs on GLPI. Into the "Computer" asset and other pages of GLPI, it always use the plural form to translate the tab names, even if the count shows only one item.

trasher: To respect parameter, sthing like that should do the trick:
```
$count = 0;
if ($_SESSION['glpishow_count_on_tabs']) {
    $count = self::getMatchingItemsCount("PluginGlpiinventoryTaskjob");
}
```

eduardomozart: Even with this it's not enough, since the ``glpishow_count_on_tabs`` is true but if there's no item on tab it still shows the string into it's singular form instead of plural.

inc/deploygroup.class.php
```diff
@@ -114,7 +114,11 @@ public function defineTabs($options = [])
         $this->addDefaultFormTab($ong);
 
         $count = self::getMatchingItemsCount("PluginGlpiinventoryTaskjob");
-        $ong[$this->getType() . '$task'] = self::createTabEntry(_n('Associated task', 'Associated tasks', Session::getPluralNumber(), 'glpiinventory'), $count);
+        if ($_SESSION['glpishow_count_on_tabs']) {
+            $tabs[2] = self::createTabEntry(_n('Associated task', 'Associated tasks', Session::getPluralNumber(), 'glpiinventory'), $count);
+        } else {
+            $tabs[2] = _n('Associated task', 'Associated tasks', Session::getPluralNumber());
+        }
```
stonebuzz, line 121:
```suggestion
        if ($_SESSION['glpishow_count_on_tabs']) {
            $count = self::getMatchingItemsCount("PluginGlpiinventoryTaskjob");
            $tabs[2] = self::createTabEntry(_n('Associated task', 'Associated tasks', Session::getPluralNumber(), 'glpiinventory'), $count);
        } else {
            $tabs[2] = _n('Associated task', 'Associated tasks', Session::getPluralNumber());
        }
```

stonebuzz: and remove `$count = self::getMatchingItemsCount("PluginGlpiinventoryTaskjob");` L116

There's no point in counting if the total is not displayed

stonebuzz (changes requested):

stonebuzz (approved):

trasher (approved):

PR #625, opened 2025-01-30
Fix PHP warnings on Deploy Package

eduardomozart:
When opening a package to edit and into another tab switch the profile to Self-Service (or any other profile which doesn't have access to Deploy Package page) it throws a PHP Warning on page if display errors is enabled on PHP settings.

## Description

image

### DIFF
--- a/inc/deploypackage.class.php
+++ b/inc/deploypackage.class.php
@@ -146,7 +146,7 @@ class PluginGlpiinventoryDeployPackage extends CommonDBTM
     {
 
         $actions = [];
-        if (strstr($_SERVER["HTTP_REFERER"], 'deploypackage.import.php')) {
+        if (isset($_SERVER["HTTP_REFERER"]) && strstr($_SERVER["HTTP_REFERER"], 'deploypackage.import.php')) {
             $actions[__CLASS__ . MassiveAction::CLASS_ACTION_SEPARATOR . 'import'] = __('Import', 'glpiinventory');
         } else {
             $actions[__CLASS__ . MassiveAction::CLASS_ACTION_SEPARATOR . 'transfert'] = __('Transfer');
@@ -166,7 +166,7 @@ class PluginGlpiinventoryDeployPackage extends CommonDBTM
     public function getForbiddenStandardMassiveAction()
     {
         $forbidden = parent::getForbiddenStandardMassiveAction();
-        if (strstr($_SERVER["HTTP_REFERER"], 'deploypackage.import.php')) {
+        if (isset($_SERVER["HTTP_REFERER"]) && strstr($_SERVER["HTTP_REFERER"], 'deploypackage.import.php')) {
             $forbidden[] = 'update';
             $forbidden[] = 'add';
             $forbidden[] = 'delete';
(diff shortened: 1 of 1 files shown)


trasher:
Displaying error from PHP config is strongly discouraged.

If user do not have access to page, I guess trying to display it is incorrect.

stonebuzz:
do you use nginx?
If yes 

Check nginx configuration, especially

  fastcgi_param  SERVER_NAME      $host;

From `/etc/nginx/fastcgi_params`

eduardomozart:
> Displaying error from PHP config is strongly discouraged.
> 
> If user do not have access to page, I guess trying to display it is incorrect.

Sorry, I think I express myself badly (my native language isn't English). I have access to the page. What happened is the following:

1. Access the Deploy Package page as Super-Admin (with proper rights).
2. Into another browser tab, change the profile to Self-Server (no proper rights anymore).
3. Change to Super-Admin again into that another browser tab.
4. Go back/reload the Deploy Package page (with proper rights).

As it expects a referrer, if it isn't set, the PHP function ``strstr()`` throws an error as the ``HTTP_REFERRER`` cannot be null (the first parameter of ``strstr`` cannot be null actually) to determine which Massive Actions to shown on Actions dropdown. This PR is innofensive and keep the current behavior of the code, just a PHP 8 fix.

trasher:
>  [...]
>  4. Go back/reload the Deploy Package page (with proper rights).

OK, sounds more legit, thanks for details.

inc/deploypackage.class.php
```diff
@@ -146,7 +146,7 @@ public function getSpecificMassiveActions($checkitem = null)
     {
 
         $actions = [];
-        if (strstr($_SERVER["HTTP_REFERER"], 'deploypackage.import.php')) {
+        if (isset($_SERVER["HTTP_REFERER"]) && strstr($_SERVER["HTTP_REFERER"], 'deploypackage.import.php')) {
```
trasher, line 149: Maybe less difficult to read, and should achieve the same result:
```suggestion
        if (strstr($_SERVER["HTTP_REFERER"] ?? '', 'deploypackage.import.php')) {
```

inc/deploypackage.class.php
```diff
@@ -166,7 +166,7 @@ public function getSpecificMassiveActions($checkitem = null)
     public function getForbiddenStandardMassiveAction()
     {
         $forbidden = parent::getForbiddenStandardMassiveAction();
-        if (strstr($_SERVER["HTTP_REFERER"], 'deploypackage.import.php')) {
+        if (isset($_SERVER["HTTP_REFERER"]) && strstr($_SERVER["HTTP_REFERER"], 'deploypackage.import.php')) {
```
trasher, line 169:
```suggestion
        if (strstr($_SERVER["HTTP_REFERER"] ?? '', 'deploypackage.import.php')) {
```

eduardomozart:
Hello @trasher, thank you for your suggestions, but your changes won't be compatible with PHP 7.4 since it accepts an empty string only from PHP 8.0 and onwards.

trasher (approved):

trasher:
As far

> Hello @trasher, thank you for your suggestions, but your changes won't be compatible with PHP 7.4 since `strstr()` accepts an empty string only from PHP 8.0 and onwards.

No, that will works. In PHP >=8, this function only accepts a string, and not the `null` value. In PHP 7.4, it accepts both.

```
 % php74 -r '$array=[]; var_dump(strstr($array["value"] ?? "", "test"));'
bool(false)
 % php82 -r '$array=[]; var_dump(strstr($array["value"] ?? "", "test"));'
bool(false)
```

eduardomozart:
I see. I commited your changes. Thank you!

PR #627, opened 2025-01-31
Improve isAgentCanDo() to check module installed flags

eduardomozart:
By default, the Deploy Package uses ``isAgentCanDo()`` function to determine which hosts to install a package, but it was also offering to install packages on ESXi remove inventory hosts which doesn't have the Deploy module installed. This PR ensures that a host which doesn't have a module installed be available to select packages to install.

## Description


stonebuzz (approved):
Seems ok for me

PR #628, opened 2025-01-31
Fix typo ('depoy' instead of 'depLoy')

eduardomozart:
Those functions seems to be called only by this file itself.

## Description

### DIFF
--- a/inc/deploypackage.class.php
+++ b/inc/deploypackage.class.php
@@ -1744,7 +1744,7 @@ class PluginGlpiinventoryDeployPackage extends CommonDBTM
         }
 
        //Get packages used for the user or a specific computer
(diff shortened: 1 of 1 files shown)


stonebuzz (approved):

trasher (approved):

trasher:
We should not rename public methods. To rename a public method, the old one that can be called from everywhere must be properly deprecated.
In the past, this plugin was required from some others, no idea if it's still the case, nor if this particular method is called.

Even if I've first accepted the PR, I wonder if we should not let incorrect methods names as is (there are plenty of them).

trasher:
Since this change does not impact at all plugin usage, I'm closing it.

PR #635, opened 2025-02-05
Align Fields and Buttons of Task to follow GLPI style guidelines

eduardomozart:

## Description

Before this PR:

image

After this PR:

image

Now the labels are end aligned to the fields, and the buttons too. The "Save" (Task management tab) and "Update" (Job configuration tab) are highlighted relative to other buttons.

### DIFF
--- a/inc/commonview.class.php
+++ b/inc/commonview.class.php
@@ -129,7 +129,7 @@ class PluginGlpiinventoryCommonView extends CommonDBTM
     public function showTextField($title, $varname)
     {
         echo "<div class='mb-2 row col-12 col-sm-6'>";
-        echo "<label class='form-label col-sm-4 col-form-label'>" . $title . "&nbsp;:</label>";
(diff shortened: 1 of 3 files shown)


eduardomozart:
> Before or after, this does not seems to reflect what is done core side

Sorry, I don't understand your question?

trasher:
Globally, when there are several buttons on a GLPI page, display looks like something like this:
image

Both display (including the "original" one) does not seems ok. 

But also I'm not a UI/UX designer.

cconard96:
> Both display (including the "original" one) does not seems ok.

Agree. If the design for these buttons is going to be changed, now is a good time to align them with the UI of GLPI itself with the buttons aligned on the right side and with the primary action button being the right-most one.

eduardomozart:
I see. I'll do it. Thank you for your feedback guys!

eduardomozart:
Done!

trasher (approved):

stonebuzz (approved):

PR #649, opened 2025-03-20
Cleanup & cs

trasher:
On way for phpstan level 5 compatible.
There are still a few fixes to be done; some on GLPI core.

First commit removes methods that seems no longer used.


stonebuzz (approved):
seems good for me

PR #659, opened 2025-04-01
Drop no longer used functions

trasher:


orthagh (commented):
I'm pretty sure that I cannot validate this.
At least, I don't sufficient time to invest in the review :)

hook.php
```diff
@@ -193,180 +193,6 @@ function plugin_glpiinventory_hook_dashboard_cards($cards)
...
-
-
-/**
- * Manage search options values
- *
- * @global object $DB
- * @param object $item
- * @return boolean
- */
-function plugin_glpiinventory_searchOptionsValues($item)
```
trasher, line 309: This one is still in use

stonebuzz (approved):

PR #682, opened 2025-05-09
Fix target list if table doesn't have is_active field

Lainow:

## Description

- It fixes !37660
Some module methods have targets that don't have the is_active field in the base, preventing the list from loading.

**Before**
image

**After**
image

### DIFF
--- a/inc/taskjobview.class.php
+++ b/inc/taskjobview.class.php
@@ -478,7 +478,10 @@ class PluginGlpiinventoryTaskjobView extends PluginGlpiinventoryCommonView
                 $condition = ['id' => $filter_id];
             }
         }
-        $condition['is_active'] = 1;
+        $itemtype_class = new $itemtype();
+        if ($DB->fieldExists($itemtype_class->getTable(), 'is_active')) {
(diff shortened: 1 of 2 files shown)


inc/taskjobview.class.php
```diff
@@ -478,7 +478,10 @@ public function ajaxModuleItemsDropdown($options)
                 $condition = ['id' => $filter_id];
             }
         }
-        $condition['is_active'] = 1;
+        $itemtype_class = new $itemtype();
+        if ($DB->fieldExists($itemtype_class->getTable(), 'is_active')) {
```
trasher, line 482: You must use  `getItemForItemtype($itemtype)` instead of `new $itemtype()`. But instanciation here is probably not necessary since the `CommonDBTM::getTable()` is static.
```suggestion
        if ($DB->fieldExists($itemtype::getTable(), 'is_active')) {
```

trasher (changes requested):
See comment

trasher (approved):

stonebuzz (commented):
Can you adapt `CHANGELOG.md`

PR #719, opened 2025-08-19
Fix phpstan issues on remaining directories

### LINKED ISSUE #713: Some path are missing in phpstan config
I think some are missing:
```suggestion
    paths:
        - ajax
        - front
        - inc
        - install
        - public
        - report
        - tests
        - hook.php
        - rector.php
        - setup.php
    excludePaths:
        - tests/files
```

_Originally posted by @Rom1-B in https://github.com/glpi-project/glpi-inventory-plugin/pull/654#discussion_r2216044766_

trasher:
close #713

### DIFF
--- a/install/update.php
+++ b/install/update.php
@@ -2133,7 +2133,7 @@ function do_profile_migration($migration)
                   AND `plugins_id`='" . $data['plugins_id'] . "'
                   AND `profiles_id`='" . $data['profiles_id'] . "'
                ORDER BY `id` DESC
-               LIMIT " . ((int)$data['cnt'] - 1) . " ";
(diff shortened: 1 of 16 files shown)


install/update.php
```diff
@@ -3933,29 +3943,6 @@ function do_networkport_migration($migration)
     );
     $migration->migrationOneTable($newTable);
 
-    // Update with mapping
```
trasher, line 3936: Alternative is to put back the `PluginGlpiinventoryMapping` class... Seems like this is useful only for very old versions of the plugin.

stonebuzz: It no longer exists on the 10bufxies branch; I can only find it on the glpi9.5 branch. In my opinion, it hasn’t been used for quite some time.

PR #745, opened 2025-09-10
Replace non existing parameter

trasher:


stonebuzz (approved):

stonebuzz:
Wee need to fix input type img too

trasher:
Weel in facts, that works on v10, no need to change ;)

PR #775, opened 2025-10-14
Fix ajax path

trasher:


trasher:
> Ok for me
> 
> There seem to be other places where _root_doc_ is missing.

Do you have examples? I take a (quick I have to admit - for`/ajax/` calls) look, but did not find any other missing

Issue #780, opened 2025-10-27
deploy - "This agent does not belong anymore in the actors defined in the job"


jluis144:
new GLPI Docker installation.

I have installed a new GLPI (Docker based) to test this error when deploying Tasks.
initially, this error have occurred in my production environment.

the error happens when trying to DEPLOY a PACKAGE to an ACTOR, if the ACTOR is a member of a GROUP_OF_COMPUTERS - if the ACTOR is a specific computer (even in that same group) the PACKAGE is DEPLOYed correctlly.

the message is --> "This agent does not belong anymore in the actors defined in the job"

<img width="488" height="341" alt="Image" src="https://github.com/user-attachments/assets/db077b1b-18f0-4768-afd8-406584a1cab4" />

thanks in advance
Jose Luis

trasher:
Seems like just the same issue than #770

Issue #770, opened 2025-10-07
GLPI plugin 1.6.0 - Error Assigning Tasks via Dynamic Groups: Agent No Longer Recognized as an Actor


root75:
### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

### Is there an existing issue for this?

- [x] I have searched the existing issues

### GLPI Version

11.0

### Plugin version

1.6.0

### Bug description

Since updating to version 1.6.0, tasks assigned via a dynamic group no longer work: I get the error "Info: This agent does not belong anymore in the actors defined in the job." However, tasks assigned by a static group still function normally

<img width="777" height="278" alt="Image" src="https://github.com/user-attachments/assets/64c2286e-b72a-41f2-8345-1fbb747c1f7d" />

### Relevant log output

```shell

```

stonebuzz:
Agent is still in actor list of task ?

It seems to be an agent flip/flop issue: when the agent comes for the first time, it gets registered in GLPI. However, once it’s assigned to the task and woken up to execute it, its record in GLPI is replaced.

trasher:
> Changes have been made on GLPI core that may affect that feature. Pleas test with a recent GLPI 11.0/bugfixes nightly build if that solves the issue.

Well.. This may not be enough. As far as we've understood; there was a problem in GLPI itself that will be fixed in 11.0.2; but a fix has also been made plugin side to prevent fatal errors that also can cause issues.

Upgrading to GLPI 11.0.2 (planned tomorrow) and plugin 1.6.1 version (planned tomorrow or the day after) will maybe fix the issue.

Mystikal91:
I'm having the same issue. 

I actually had the "No active session" problem and while upgrading to 1.6.2 solved it, it didn't solve the "Agent No Longer Recognized as an Actor, so #800 didn't solved it for me

trasher:
We need information to reproduce that issue on a fresh install in order to fix; I have no idea what's wrong right now.

root75:
Same problem for me again with a new agent on a fresh install. I'm now using static groups for the moment. What information should I send to help with the debugging?

stonebuzz:
Hi @root75 , @ka0tik0, @ka0tik0 

can you try this -> https://github.com/glpi-project/glpi-inventory-plugin/pull/812

Best regards

PR #785, opened 2025-11-03
Fix dynamic groups actors

trasher:
Since https://github.com/glpi-project/glpi/pull/21622 Visibility is less restrictive depending on session interface that was previously not used on that part

Maybe session interface should be changed at an higher level, but I'm afraid this causes other issues in the plugin.
Any better solution would be welcome.


AdrienClairembault (approved):

trasher:
> @trasher can you adat `CHANGELOG.md` ?

done

trasher:
Waiting for feedback on #780 and #770 before merging

PR #801, opened 2025-11-10
Fix for adding targets and actors to a task

Lainow:

## Description

- It fixes !40236
No action occurs when attempting to add a target or actor to a task in GLPI inventory.

glpiinv

### DIFF
--- a/inc/deploypackage_group.class.php
+++ b/inc/deploypackage_group.class.php
@@ -67,7 +67,7 @@ class PluginGlpiinventoryDeployPackage_Group extends CommonDBRelation
     /**
      * Set we don't check parent right of the second item
      *
-     * @var 1|2|3
+     * @var self::DONT_CHECK_ITEM_RIGHTS|self::HAVE_VIEW_RIGHT_ON_ITEM|self::HAVE_SAME_RIGHT_ON_ITEM
      */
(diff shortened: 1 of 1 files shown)


inc/taskjobview.class.php
```diff
@@ -46,6 +46,10 @@ class PluginGlpiinventoryTaskjobView extends PluginGlpiinventoryCommonView
      */
     public function __construct()
     {
+        /** @var array $CFG_GLPI */
+        global $CFG_GLPI;
+        $base_url = $CFG_GLPI['url_base'];
+
```
trasher: Not used (but if a fix must be done, it's probably in `getBaseUrlFor()` method).

inc/deploypackage_entity.class.php
```diff
@@ -67,7 +67,7 @@ class PluginGlpiinventoryDeployPackage_Entity extends CommonDBRelation
     /**
      * Set we don't check parent right of the second item
      *
-     * @var integer
+     * @var 1|2|3
```
trasher:
```suggestion
     * @var self::DONT_CHECK_ITEM_RIGHTS|self::HAVE_VIEW_RIGHT_ON_ITEM|self::HAVE_SAME_RIGHT_ON_ITEM
```

PR #823, opened 2025-11-20
Use constants to prevent miunderstanding

trasher:
Seems like other occurrences (in that class) already are using constants.

### DIFF
--- a/inc/task.class.php
+++ b/inc/task.class.php
@@ -711,7 +711,7 @@ class PluginGlpiinventoryTask extends PluginGlpiinventoryTaskView
                 'state.state' => [
                     PluginGlpiinventoryTaskjobstate::FINISHED,
                     PluginGlpiinventoryTaskjobstate::IN_ERROR,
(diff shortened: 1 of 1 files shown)


inc/task.class.php
```diff
@@ -708,7 +708,11 @@ public function cleanTasksAndJobs($interval)
             'WHERE' => [
                 'task.is_deploy_on_demand' => 1,
                 new QueryExpression('DATEDIFF(ADDDATE(log.date, INTERVAL ' . (int) $interval . ' DAY), CURDATE()) < 0'),
-                'state.state' => [3, 4, 5],
+                'state.state' => [
+                    PluginGlpiinventoryTaskjobstate::FINISHED,
+                    PluginGlpiinventoryTaskjobstate::IN_ERROR,
+                    PluginGlpiinventoryTaskjobstate::CANCELLED
```
trasher:
```suggestion
                    PluginGlpiinventoryTaskjobstate::CANCELLED,
```

PR #866, opened 2026-01-06
CS

trasher:
Update cs-fixer configuration
Update Rector configuration
Fix wrong condition in build file


trasher:
I did not pay attention fixer had errors trying to fix :/ And I do not understand the problem for now.

PR #868, opened 2026-01-07
Handle constants the same way core does 

trasher:
First commit is constants rplacement, second a phpstan fix (no idea why it's was not seen before)

### DIFF
--- a/setup.php
+++ b/setup.php
@@ -341,7 +341,7 @@ function plugin_init_glpiinventory()
             }
         }
     } else { // plugin not active, need $moduleId for uninstall check
-        include_once(PLUGIN_GLPI_INVENTORY_DIR . '/inc/module.class.php');
+        include_once(__DIR__ . '/inc/module.class.php');
     }
 
     // exclude some pages from splitted layout


setup.php
```diff
@@ -38,37 +38,26 @@
...
+    'PLUGIN_GLPI_INVENTORY_GLPI_MIN_VERSION' => '11.0.2', // Minimal GLPI version, inclusive
+    'PLUGIN_GLPI_INVENTORY_GLPI_MAX_VERSION' => '11.0.99', // Maximum GLPI version, exclusive
+    'PLUGIN_GLPI_INVENTORY_DIR' => __DIR__,
+    'PLUGIN_GLPI_INVENTORY_OFFICIAL_RELEASE' => 0,
+    'PLUGIN_GLPI_INVENTORY_REPOSITORY_DIR' => GLPI_PLUGIN_DOC_DIR . '/glpiinventory/files/repository/',
+    'PLUGIN_GLPI_INVENTORY_MANIFESTS_DIR' => GLPI_PLUGIN_DOC_DIR . '/glpiinventory/files/manifests/',
+    'PLUGIN_GLPI_INVENTORY_UPLOAD_DIR' => GLPI_PLUGIN_DOC_DIR . '/glpiinventory/upload/',
+];
+foreach ($constants as $name => $default_value) {
+    if (!defined($name)) {
```
cedric-anne: It will make them customizable, but I guess noone will be stupid enough to define them in its local_define.php file.

trasher: That's not the goal, I get errors with phpstan saying constants already are defined if I don't check defined :/

PR #873, opened 2026-01-12
Drop no longer used methods

trasher:
`Taskjob::prepareRunTaskjob()` was the only place where those were called; and it's been removed since it was no longer called (https://github.com/glpi-project/glpi-inventory-plugin/pull/871/commits/4666a063394491d2a5e7590c8687cfd950245c25).

I've tested with collect; all seems OK; so I guess this code was dead from a while.


trasher:
> Tested for netdiscovery task: this is also not used.
> 
> I agree these methods should indeed be dead code since years.

Great, thank you for testing :+1:

Rom1-B (approved):

PR #875, opened 2026-01-13
Fix html link displayed; closes #874

### LINKED ISSUE #874: Seems link in task job report is again rendered as plain text in place of html link
### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

### Is there an existing issue for this?

- [x] I have searched the existing issues

### GLPI Version

11.0.4

### Plugin version

1.6.5

### Bug description

I know such an issue was fixed few timed ago, but maybe it was back after some later work:

Image

I expect to see a link in place of plain text html source and being able to directly access the asset via the link if I need it.

### Relevant log output

```shell

```

### Page URL

_No response_

### Steps To reproduce

1. Create a netdiscovery task and eventually a netinventory one
2. Run the task as usual, be sure to have at least one device in the targeted ip ramge
3. Check the task run log, the link for imported device is output as plain text html code

### Your GLPI setup information

_No response_

### Anything else?

Also noticed it when using 1.6.4 last week.

I'm using FF 146.0 on linux as browser.

trasher:


trasher:
See https://github.com/glpi-project/glpi-inventory-plugin/issues/876

PR #877, opened 2026-01-14
Drop unused

trasher:
To be reviewed per commit


stonebuzz (approved):
Sounds good, we’ll see how it goes.

PR #879, opened 2026-01-15
Replace some hardcoded class names

trasher:


inc/collectcontentcommon.class.php
```diff
@@ -102,28 +102,26 @@ public function getTabNameForItem(CommonGLPI $item, $withtemplate = 0)
...
+                    $fk = getForeignKeyFieldForItemType($collect);
+                    if (
+                        ($nb = countElementsInTable(
+                            $this->getTable(),
+                            [$fk => $in]
+                        )) > 0
+                    ) {
+                        return self::createTabEntry($collect::getTypeName(Session::getPluralNumber()), $nb, null, $class::getIcon());
+                    }
+                }
```
Rom1-B:
```suggestion
                }
```

report/ports_date_connections.php
```diff
@@ -193,7 +193,7 @@ function displaySearchForm(): void
     $_SERVER["REQUEST_URI"] = buildSavedSearchUrl($_SERVER["REQUEST_URI"], $_GET);
     TemplateRenderer::getInstance()->render('pages/tools/savedsearch/save_button.html.twig', [
         'type' => SavedSearch::SEARCH,
-        'itemtype' => 'PluginGlpiinventoryNetworkport2',
+        'itemtype' => 'PluginGlpiinventoryNetworkport2', //FIXME?
```
Rom1-B: Can you be more specific?

trasher: No :D
This should be an itemtype, but this class does not exists. It's been like that for years; and I have no idea why :/

trasher: The idea is to mark here there is something wrong, I honestly do not really want to investigate too much; this PR was supposed to be simple ;)

PR #886, opened 2026-01-19
Use DbTestCase and ensure DB transaction is not commited

trasher:
Also some minor changes on code, see each commit.

DB transaction in tests has been done here only on the `Unit` directory.
I've begin to work on the `Integration` one, but there are many DB dependencies including between tests files; that's way more complex.

### DIFF
--- a/inc/communicationrest.class.php
+++ b/inc/communicationrest.class.php
@@ -91,23 +91,23 @@ class PluginGlpiinventoryCommunicationRest
             $a_agent       = $params['agent']->fields;
 
             foreach (array_keys($params['task']) as $task) {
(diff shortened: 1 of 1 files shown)


tests/Unit/TimeslotTest.php
```diff
@@ -304,7 +305,7 @@ public function testAddEntryIn3Ranges()
 
     public function testAddEntryForTwoDays()
     {
-
+        $this->testAddEntryIn3Ranges();
```
trasher: Maybe should I take a look to propose a better fix here too (may be less simple).

trasher: OK, I won't. I do not understand those tests; they probably should be correctly rewritten, this is clrearly out of the PR scope.

PR #930, opened 2026-04-10
Update communicationnetworkinventory.class.php

JohTechTHD:
Checklist before requesting a review

  - I have performed a self-review of my code.
  - I have added tests (when available) that prove my fix is effective or that my feature works.
  - I have updated the CHANGELOG with a short functional description of the fix or new feature.
  - This change requires a documentation update.

  Description

  - It fixes # (not applicable — new feature)
  - Adds a glpiinventory_post_network_inventory hook that fires in
  PluginGlpiinventoryCommunicationNetworkInventory::import() after doInventory() completes successfully, providing
  third-party plugins with both the persisted asset item and the full raw SNMP inventory data.

  Problem:
  There is currently no hook in the glpiinventory code path that fires after the network device has been written to the
  database. The core GLPI NETWORK_INVENTORY hook fires before doInventory(), so newly discovered devices may not exist
  in the DB yet. The core GLPI POST_INVENTORY hook does not fire on the glpiinventory path at all.

  This makes it impossible for third-party plugins to reliably post-process network inventory data (e.g. topology
  linking, module asset creation, impact relation building) when devices are managed via glpiinventory agents.

  Solution:
  One line added in the else branch (inventory accepted, not refused) after doInventory() succeeds:

  Plugin::doHookFunction('glpiinventory_post_network_inventory', [
      'item'     => $item,      // asset already persisted in DB
      'raw_data' => $a_CONTENT, // full SNMP data (network_ports, network_components, …)
  ]);

  Third-party plugins register via:
  $PLUGIN_HOOKS['glpiinventory_post_network_inventory']['myplugin']
      = 'plugin_myplugin_callback';

  The change is purely additive — no existing behavior is modified.

### DIFF
--- a/inc/communicationnetworkdiscovery.class.php
+++ b/inc/communicationnetworkdiscovery.class.php
@@ -159,8 +159,8 @@ class PluginGlpiinventoryCommunicationNetworkDiscovery
                                 . ' [[' . $item::class . '::' . $item->getID() . ']]';
                             $this->addtaskjoblog();
                             Plugin::doHookFunction('glpiinventory_post_network_discovery', [
-								'item'     => $item,
-								'raw_data' => $a_CONTENT,
+                                'item'     => $item,
+                                'raw_data' => $a_CONTENT,
 							]);
                         }
                         $response = ['response' => ['RESPONSE' => 'SEND']];
--- a/inc/communicationnetworkdiscovery.class.php
+++ b/inc/communicationnetworkdiscovery.class.php
@@ -158,7 +158,7 @@ class PluginGlpiinventoryCommunicationNetworkDiscovery
                                 = '[==detail==] ' . $what . ' ' . $item->getTypeName()
                                 . ' [[' . $item::class . '::' . $item->getID() . ']]';
                             $this->addtaskjoblog();
-							Plugin::doHookFunction('glpiinventory_post_network_discovery', [
+                            Plugin::doHookFunction('glpiinventory_post_network_discovery', [
 								'item'     => $item,
 								'raw_data' => $a_CONTENT,
 							]);
--- a/inc/communicationnetworkdiscovery.class.php
+++ b/inc/communicationnetworkdiscovery.class.php
@@ -158,6 +158,10 @@ class PluginGlpiinventoryCommunicationNetworkDiscovery
                                 = '[==detail==] ' . $what . ' ' . $item->getTypeName()
                                 . ' [[' . $item::class . '::' . $item->getID() . ']]';
                             $this->addtaskjoblog();
+							Plugin::doHookFunction('glpiinventory_post_network_discovery', [
+								'item'     => $item,
+								'raw_data' => $a_CONTENT,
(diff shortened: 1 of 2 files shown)


g-bougard (changes requested):
You changed indentation where you shouldn't have to. Then in place of just adding few lines, you seem to have changed more.

Please restore indentation so it is also coherent with the true case implemented before the `} else {`.

JohTechTHD:
Thank you for the review!

You are absolutely right. I accidentally changed the indentation in surrounding code while adding the hook lines. I will restore the original indentation and only add the minimal required lines, keeping the style consistent with the existing code.

trasher:
I'm really not sure plugins should add their own hooks. 
And that's probably not a good idea neither to rely on inventory plugin as a dependency.

JohTechTHD:
Hi @trasher, thank you for the feedback 

I understand the hesitation around plugins adding their own hooks and the dependency issue. Let me explain the concrete problem I'm trying to solve, maybe that helps put it in perspective.

We run Cisco switches (stacked) with pluggable modules. These modules are independently purchased assets and should ideally be tracked separately in inventory/CMDB — but right now there's no reliable way for a third-party plugin to act on the raw SNMP data *after* the network device has been persisted to the database. The same goes for CDP neighbor linking  use cases.

 So currently, there is simply no extension point available for post-processing.

My alternatives are:
1. Write a completely independent SNMP inventory plugin that doesn't use glpiinventory at all — which would mean duplicating a lot of work and losing all the value glpiinventory already provides.
2. Or add this small, purely additive hook that gives any third-party plugin a clean, reliable extension point without changing any existing behavior.

I genuinely believe option 2 is the better outcome for the ecosystem. 

I'm open to discussing the right approach. If there's a preferred pattern for this kind of extensibility in glpiinventory, I'm happy to follow it. But I do think the missing extension point is a real gap worth addressing.

trasher (approved):

trasher:
Please fix lint issues

trasher (changes requested):
It's OK for me. Please fix lint before we can merge.

inc/communicationnetworkdiscovery.class.php
```diff
@@ -158,6 +158,10 @@ public function import($p_DEVICEID, $a_CONTENT, Inventory $inventory): array
                                 = '[==detail==] ' . $what . ' ' . $item->getTypeName()
                                 . ' [[' . $item::class . '::' . $item->getID() . ']]';
                             $this->addtaskjoblog();
+                            Plugin::doHookFunction('glpiinventory_post_network_discovery', [
+								'item'     => $item,
+								'raw_data' => $a_CONTENT,
+							]);
```
g-bougard: I still see tabulation chars there. Please use spaces.

trasher:
lint is still failing

trasher:
lint is still broken. Please run php-cs-fixer.

trasher (approved):

PR #963, opened 2026-06-30
fix (import) produce a valid CSV when exporting task execution results

Herafia:

## Description

- It fixes !44976
- The export used to generate an unusable file: the rows were misaligned, and the "last log" column was generated without proper escaping (semicolons, quotation marks, line breaks, etc.).
We now use phpsreadsheet to ensure compliance and generate the file correctly.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [UNRELEASED]
 
-- Fix export csv
+- Fix task execution results CSV export
 
 ## [1.6.8] - 2026-06-24
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,10 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
+## [UNRELEASED]
+
+- Fix export csv
+
 ## [1.6.8] - 2026-06-24
 
 - Fix module / exception handling during updates from the agent or general settings.
--- a/inc/taskview.class.php
+++ b/inc/taskview.class.php
@@ -551,6 +551,8 @@ class PluginGlpiinventoryTaskView extends PluginGlpiinventoryCommonView
      */
     public function csvExport($params = [])
     {
+        global $CFG_GLPI;
+
         $default_params = [
             'agent_state_types' => [],
         ];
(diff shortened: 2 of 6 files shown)


trasher:
Why not relying on core CSV exports (phpsreadsheet)?

CHANGELOG.md
```diff
@@ -5,6 +5,10 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
+## [UNRELEASED]
+
+- Fix export csv
```
stonebuzz:
```suggestion
- Fix task execution results CSV export
```

tests/Unit/TaskCsvExportTest.php
```diff
@@ -0,0 +1,160 @@
...
+    {
+        return ['tasks' => [1 => [
+            'task_name' => 'Test task',
+            'jobs'      => [10 => [
+                'name'    => 'Test job',
+                'method'  => 'inventory',
+                'targets' => [99 => [
+                    'name'   => 'Test target',
+                    // agent_id = 0 → Agent::getFromDB() fails, so the name stays empty
+                    'agents' => [0 => $execs],
```
stonebuzz, line 59: All test cases set `agent_id = 0` so that `Agent::getFromDB(0)` always fails, leaving columns 4 (Agent) and 5 (Computer name) always empty. Can you add a test case that creates a real `Agent` and `Computer` row in the DB and asserts those two columns are populated correctly?

stonebuzz (changes requested):

trasher:
(also, a note: rector is not ran on this plugin, we may add that but that's out of the scope of this PR ;))

inc/taskview.class.php
```diff
@@ -592,92 +593,111 @@ public function csvExport($params = [])
...
+        $writer
+            ->setDelimiter($_SESSION['glpicsv_delimiter'] ?? ';')
+            ->setEnclosure('"')
+            ->setUseBOM(true)
+            ->setLineEnding("\r\n")
+            ->setSheetIndex(0);
+        $writer->save('php://output');
+
+        // force exit to prevent further display
+        exit; //@phpstan-ignore-line (whole method needs to be refactored)
```
Rom1-B, line 611: Are you doing this refactoring in another PR?

inc/taskview.class.php
```diff
@@ -592,92 +593,111 @@ public function csvExport($params = [])
...
-                    if (!$last($task['jobs'], $job_id)) {
-                        echo SEP;
-                    }
+                        foreach ($target['agents'] as $agent_id => $agent) {
+                            $agent_obj->getFromDB($agent_id);
+                            $agent_name = $agent_obj->getName();
+                            $computer_name = '';
+                            if (isset($agent_obj->fields['items_id']) && $computer->getFromDB($agent_obj->fields['items_id'])) {
+                                $computer_name = $computer->getName();
+                            }
```
Rom1-B: `$agent_obj->getFromDB()` return value is not checked. When an agent no longer exists in the database the call returns `false` but leaves `$agent_obj->fields` populated with the **previous** iteration's data, so the stale agent name (and its computer) are written into the CSV row.

```suggestion
                            $agent_name = '';
                            $computer_name = '';
                            if ($agent_obj->getFromDB($agent_id)) {
                                $agent_name = $agent_obj->getName();
                                if (isset($agent_obj->fields['items_id']) && $computer->getFromDB($agent_obj->fields['items_id'])) {
                                    $computer_name = $computer->getName();
                                }
                            }
```

Rom1-B (changes requested):

inc/taskview.class.php
trasher: As far as I remember, fixing this requires to use Controllers; that's not something planned in this plugin right now.

Rom1-B (approved):

trasher (approved):
lgtm

stonebuzz (approved):
LGTM

PR #971, opened 2026-07-02
GLPI v12

trasher:


trasher:
I've opened a PR to fi issue on CI (https://github.com/glpi-project/glpi/pull/25404)

PR #978, opened 2026-07-09
Update timeslot UI

trasher:
Goal was to migrate this part to Twig; the interface is quite similar as it was previously.

I've also replaced the weekly view with a simple HTML table (no need for a js lib here).


stonebuzz (approved):
LGTM.

Tested and verified functionally.

inc/timeslotentry.class.php
```diff
@@ -193,83 +143,62 @@ public function formDeleteEntry($timeslots_id): void
...
+     * Each day is split into 96 quarter-hour cells; a cell is active when it falls within
+     * a configured entry range [begin, end[ (boundaries expressed in seconds).
      *
-     * @param int $timeslots_id
+     * @param array<int,string> $days           Days of week keyed by ISO numeric value.
+     * @param array<int,array<array<string,mixed>>> $entries_by_day Entries grouped by day.
+     * @return array<array{day:int,label:string,cells:array<bool>}>
      */
-    public function showTimeSlot(int $timeslots_id): void
+    private function getWeekGrid(array $days, array $entries_by_day): array
```
Rom1-B, line 179: `getWeekGrid()` is the core logic of this migration (converts DB ranges into 96 cells/day) but is untested. `TimeslotTest.php` only covers `addEntry()` (unchanged). A boundary bug (e.g. `end == 86400`) would go unnoticed.

can you add a test with entries spanning a full day and midday boundaries?

Rom1-B (changes requested):
`inc/toolbox.class.php:273`

`PluginGlpiinventoryToolbox::showHours()` is no longer called anywhere in the plugin (replaced by the `$hours` array + `Dropdown::showFromArray` in Twig). It's public, so it could potentially be used outside the plugin, but otherwise, it's dead code.

inc/timeslotentry.class.php
trasher: Tests have been added

trasher:
> `inc/toolbox.class.php:273`
> 
> `PluginGlpiinventoryToolbox::showHours()` is no longer called anywhere in the plugin (replaced by the `$hours` array + `Dropdown::showFromArray` in Twig). It's public, so it could potentially be used outside the plugin, but otherwise, it's dead code.

Indeed; I removed that method

Rom1-B (approved):

PR #979, opened 2026-07-10
Fix: remove dead blacklist migration on plugin update

Rom1-B:

## Description

- It fixes !45218
- On update, the plugin recreated legacy blacklist tables and repopulated their default rows, only to have those same tables dropped a few steps later in the same migration chain.
- On instances with drifted ids in those tables (old FusionInventory history, custom criteria), one of the hardcoded inserts collided with an existing primary key and made the update fail with a duplicate entry error, blocking it permanently since the migration ran again on every retry.
- The dead migration is now removed entirely; only its one real side effect (a display preference rename) is kept.


ajax/subvisibility.php
trasher, line 1: This file was not part of FI plugin

ajax/visibility.php
trasher, line 1: This file was not part of FI plugin

ajax/subvisibility.php
trasher: (same most of files this header was added)

Rom1-B: removed

ajax/visibility.php
Rom1-B: removed

trasher (approved):

trasher:
Please rebase to solve conflict ;)

stonebuzz (approved):

stonebuzz:
image

trasher:
> image

It's OK on my side; this is the kind of thing that happen on gh ;)

PR #982, opened 2026-07-10
Remove non existing Safe function

trasher:


trasher:
After check, the function seem to be present in Safe; so it should not give an error.

stonebuzz:
Yes, it's present, but it was not introduced in GLPI 11.0.8. 

I assume you're using the bugfixes branch/version, which already includes this change.

PR #974, opened 2026-07-06
Updated registry collect

g-bougard:
## Checklist before requesting a review

- [x] I have performed a self-review of my code.
- [x] I have added tests (when available) that prove my fix is effective or that my feature works.
- [x] I have updated the CHANGELOG with a short functional description of the fix or new feature.

## Description

- It fixes !42477
- It completes and replaces #964 as @Lainow is in vacation
- Here is a brief description of what this PR does:

Add registry collect modes: Check path existence and Check if key is defined and Add a Recursion depth option to the registry collect to read sub-keys recursively.

PR #988, opened 2026-07-28
fix(collect): handle REG_MULTI_SZ and hex DWORDs consistently

SerWax:

- [ ] I have updated the CHANGELOG with a short functional description of the fix or new feature.

## Description

updateComputer() receives the agent's registry data and treats it as
scalar strings on both its update and insert paths. Two value types
break that assumption.

A REG_MULTI_SZ value is decoded by the agent as a list, so $value is an
array. The insert path calls preg_match() on it and raises

  TypeError: Safe\preg_match(): Argument #2 ($subject) must be of type
  string, array given, called in collect_registry_content.class.php:134

which is uncaught, so the endpoint answers 500 and the agent discards
its whole submission. Agent-side this appears as "Bad answer: CSRF
checking is failing", which is misleading: no CSRF check is involved and
nothing is written to GLPI's access-errors.log. The update path fails
silently instead, storing the literal string "Array". One multi-string
in a collected key is enough to lose every other value in that Collect.

A REG_DWORD arrives as a hex string. The insert path converts it with
hexdec(), the update path does not, so a value is decimal on its first
harvest and "0x0000002D" on every harvest after. A consumer casting that
to int gets 0, and cannot tell the two encodings apart.

Both are fixed where the data enters: multi-strings are joined once,
immediately after _sid is removed, which covers both paths, and the
existing hex conversion is applied on the update path too.

Tested on GLPI 11.0.8, GLPI Inventory 1.6.8, GLPI Agent 1.18, Windows
hosts, with a Collect of seven keys mixing REG_SZ, REG_DWORD and
REG_MULTI_SZ values. Before: four keys returned nothing and the task
reported an error. After: all keys harvest and values store decimal.

### DIFF
--- a/inc/collect_registry_content.class.php
+++ b/inc/collect_registry_content.class.php
@@ -111,9 +111,17 @@ class PluginGlpiinventoryCollect_Registry_Content extends PluginGlpiinventoryCol
         }
 
         unset($registry_data['_sid']);
+        foreach ($registry_data as $k => $v) {
+            if (is_array($v)) {
+                $registry_data[$k] = implode("\n", $v);
+            }
+        }
         foreach ($registry_data as $key => $value) {
             foreach ($db_registries as $keydb => $arraydb) {
                 if ($arraydb['key'] == $key) {
+                    if (preg_match("/^0x[0-9a-fA-F]{1,}$/", $value)) {
+                        $value = hexdec($value);
+                    }
                     $input = ['key'   => $arraydb['key'],
                         'id'    => $keydb,
                         'value' => $value,


trasher:
I do not use Windows nor have any skills on it; I can't tell if the fix is correct.
Anyway, this probably needs a test case.

g-bougard:
Hi @SerWax 

Actually, we have a dev for a customer which may be in conflict: #974
And agent-side, we have also some changes here: glpi-project/glpi-agent#1201

On agent-side, if I remember well, this will changes the way arrays are sent back to server. So your PR may be not correct in that case.

Our dev is actually in review by our customer and will be merged after eventually more changes.

You can eventually still test glpi-agent changes using the development version you'll find in the Windows-x64 artifact from the GH Actions of the PR packaging build: https://github.com/glpi-project/glpi-agent/actions/runs/28798940152/artifacts/8112050542

So I propose anyway to wait for that dev to be included to be sure we don't mess up something. You still can add test cases as suggested by @trasher as it would help to keep result coherent other time.

SerWax:
After reviewing I agree that there is probable conflict. I'd wait for the final release of the new inventory plugin and agent, before testing and verifying if my change shall still apply and how.

PR #993, opened 2026-08-06
Fix: harmonize rights check on deploy package content edition

Rom1-B:

## Description

- It fixes #N/A
- Rights checks on deploy package content edition (actions, files, checks) were inconsistent across entry points.
- All entry points now rely on the same rights check already used for the package edition UI.


trasher (approved):
LGTM from a code point of view, I cannot test

g-bougard (approved):
Seems good for me.

PR #1018, opened 2026-09-18
Fix: validate submitted selection on self-service deployment

stonebuzz:

## Description

- It fixes #N/A
- The self-service deployment page listed only the packages and computers offered to the connected user, but the form submission was processed without checking that the submitted values belonged to that list.
- The submitted computer and package identifiers are now compared against the offering computed for the current user, and the target computer must belong to an entity the user can access. Anything else is ignored.


Rom1-B (approved):

front/deploypackage.public.php
```diff
@@ -45,42 +48,56 @@
...
+    }
+
+    //Every computer having received a package must have its agent woken up
+    $deployed_computers_ids = array_keys($deployments);
+
+    $agent_rows = [];
+    if ($deployed_computers_ids !== []) {
+        foreach ($DB->request([
+            'FROM'   => Agent::getTable(),
+            'WHERE'  => ['itemtype' => 'Computer', 'items_id' => $deployed_computers_ids],
```
trasher:
```suggestion
            'WHERE'  => ['itemtype' => Computer::class, 'items_id' => $deployed_computers_ids],
```

trasher (approved):
Code seem correct, but I cannot test

